PR #6, opened 2022-01-07
health check and tests

lbennett-moj:


src/main/resources/application-dev.yml
```diff
@@ -0,0 +1,14 @@
+server:
+  shutdown: immediate
+
+api.base.url:
+  nomis: http://localhost:8081
+  oauth: http://localhost:8090/auth
+
+
+spring:
```
andymarke, line 9: I think this is not needed

andymarke (approved):
LGTM

I think we need to docker-compose next

PR #14, opened 2022-01-12
Initial integration test for consumption of visits message

lbennett-moj:


andymarke (approved):
LGTM

I think with the bash files we can have 2 sets; one that we commit that work against localstack and then think about how we can handle the T3 version later

PR #17, opened 2022-01-13
Add integration to a future Visits API

andymarke:
For now we will mock that out by pointing back to this service


helm_deploy/values-dev.yaml
```diff
@@ -12,6 +12,8 @@ generic-service:
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.dev.json
     API_BASE_URL_OAUTH: https://sign-in-dev.hmpps.service.justice.gov.uk/auth
     API_BASE_URL_NOMIS: https://nomis-prsner-dev.aks-dev-1.studio-hosting.service.justice.gov.uk
+    # Set to ourselves until new service is ready
```
andymarke, line 15: since real API is only partially developed we will point back to this service and create temporary stub APIs

hyderkhanmoj (approved):

PR #33, opened 2022-01-24
Sdi 26 cancel message

lbennett-moj:


.gitignore
```diff
@@ -70,3 +70,4 @@ ecr.repo
 #Helm
 **/Chart.lock
 /send-visit-create-message-sns.bash
+/send-visit-cancel-message-sns.bash
```
petergphillips, line 73: wondering if these scripts could be checked in instead?

petergphillips (approved):
LGTM

.gitignore
lbennett-moj: yes, i think we need a better way, will look at checking in a template without tokens etc

PR #36, opened 2022-01-27
SDI-41 handle conflict response from nomis api

lbennett-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/NomisApiServiceTest.kt
```diff
@@ -54,22 +53,6 @@ internal class NomisApiServiceTest {
...
-                  }
-     """
-      )
-
-      val result = nomisApiService.createVisit(newVisit())
-
-      assertThat(result.visitId).isEqualTo("12345")
-    }
-
     @Test
```
lbennett-moj, line 56: currently(!) the mapping is on the nomis side so response not of interest

andymarke (approved):
LGTM

PR #41, opened 2022-02-11
SDI-58 Implement id mapping in the synchronisation service

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/WebClientConfiguration.kt
```diff
@@ -12,29 +12,39 @@ import org.springframework.security.oauth2.client.web.reactive.function.client.S
...
+      .apply(oauth2Client.oauth2Configuration())
       .build()
   }
 
   @Bean
-  fun nomisApiWebClient(authorizedClientManager: OAuth2AuthorizedClientManager): WebClient {
+  fun mappingWebClient(authorizedClientManager: OAuth2AuthorizedClientManager): WebClient {
     val oauth2Client = ServletOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager)
-    oauth2Client.setDefaultClientRegistrationId("nomis-api")
+    oauth2Client.setDefaultClientRegistrationId("system-api")
```
steverendell, line 44: I thought it makes sense for mapping to share the client id wtih nomis api, which means it needs the READ_NOMIS role adding (which Ive done in dev)

andymarke: Actually, our convention is a set of client credentials per system a service connects to: so for instance the migration service uses 3 sets; one for nomis api, one for VSIP and one for mapping service.

The Auth team are aware of that pattern and are happy to create new credentials for each environment so suggest requesting that and changing back to `nomis-api` and add a new client registration

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/NomisApiService.kt
```diff
@@ -19,27 +19,22 @@ class NomisApiService(@Qualifier("nomisApiWebClient") private val webClient: Web
...
     webClient.post()
       .uri("/prisoners/${request.offenderNo}/visits")
       .bodyValue(request)
       .retrieve()
-      .bodyToMono(Unit::class.java)
-      .onErrorResume(WebClientResponseException.Conflict::class.java) {
-        log.warn("createVisit failed for offender no ${request.offenderNo} and vsip visit id ${request.vsipVisitId}  with message ${it.message}")
-        Mono.empty()
-      }
+      .bodyToMono(String::class.java)
```
andymarke, line 27: Does "create" just return a string? I would have thought it would return an object for the new visit, at least an object with a visit id in rather just a non-JSON string.

That is the pattern we are using for creating VSIP visits - tends to be more flexible to return an JSON object that a primitive String

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/MappingService.kt
```diff
@@ -0,0 +1,69 @@
...
+
+  val nomisId: String,
+
+  val vsipId: String,
+
+  @Schema(description = "Label (a timestamp for migrated ids)")
+  @Size(max = 20)
+  val label: String? = null,
+
+  @Schema(description = "Mapping type", allowableValues = ["MIGRATED", "ONLINE"], required = true)
```
andymarke, line 65: Is that Schema used anywhere? Not sure this ever appears in swagger does it?

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/visits/PrisonVisitsService.kt
```diff
@@ -1,61 +1,107 @@
...
+            endTime = this.endTime,
+            visitorPersonIds = this.visitors.map { it -> it.nomisPersonId },
+            issueDate = visitBookedEvent.bookingDate,
+            visitType = "SCON", // TODO mapping
+          )
+        )
+      } catch (e: Exception) {
+        telemetryClient.trackEvent("visit-booked-create-failed", telemetryMap)
+        log.warn("Unexpected exception", e)
+        return
```
andymarke, line 63: doesn't this *have* to re throw ? Else we lose this booking for ever, and it won't go in DLQ?

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/visits/PrisonVisitsService.kt
```diff
@@ -1,61 +1,107 @@
...
+      val mapWithNomisId = telemetryMap.plus(Pair("nomisVisitId", nomisId!!))
+
+      try {
+        mappingService.createMapping(
+          MappingDto(nomisId = nomisId, vsipId = visitBookedEvent.visitId, mappingType = "ONLINE")
+        )
+      } catch (e: Exception) {
+        telemetryClient.trackEvent("visit-booked-create-map-failed", mapWithNomisId)
+        log.warn("Unexpected exception", e)
+        return
```
andymarke, line 75: yep, makes sense not to go on DLQ (very minor I suspect that should be logged as error)

andymarke (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/MappingService.kt
```diff
@@ -0,0 +1,69 @@
...
+  fun createMapping(request: MappingDto) {
+    webClient.post()
+      .uri("/mapping")
+      .bodyValue(request)
+      .retrieve()
+      .bodyToMono(Unit::class.java)
+      .block()
+  }
+
+  fun getMappingGivenNomisId(nomisId: Long): MappingDto? =
```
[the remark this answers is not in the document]
steverendell: It doesnt. I wasnt sure whether to include it ... as it stands it makes the MappingService 'complete' but could remove?

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/NomisApiService.kt
steverendell: Correct - it does return an object, will fix

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/visits/PrisonVisitsService.kt
steverendell: 👍

andymarke (approved):
LGTM

PR #110, opened 2022-09-28
Switch off incentives synchronisation in live

andymarke:


steverendell (commented):
looking at the code im not sure thisll work as the event type is 
incentives.iep-review.inserted
and so the property is
feature.event.incentives.iep-review.inserted
and I dont know if that hyphen gets translated to an underscore

steverendell (approved):

andymarke:
> looking at the code im not sure thisll work as the event type is
> incentives.iep-review.inserted
> and so the property is
> feature.event.incentives.iep-review.inserted
> and I dont know if that hyphen gets translated to an underscore

Actually this is exact variable I added to the live pod and seems to work:

In live we now get `Feature switch is disabled for incentives.iep-review.inserted`  

In other words the hyphon gets converted to underscore

PR #131, opened 2022-12-01
SDI-336 Improved detection of IEP created in NOMIS

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesService.kt
```diff
@@ -84,7 +85,7 @@ class IncentivesService(
         return
       }
 
-      telemetryClient.trackEvent("incentive-event", mapWithNomisId)
+      telemetryClient.trackEvent("incentive-created-event", mapWithNomisId)
     }
```
andymarke, line 89: slightly better name for this :-)

lbennett-moj (approved):
LGTM

PR #147, opened 2023-01-18
SDI-528 Added integration to (non-existent) sentence adjustment API

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/health/HealthCheck.kt
```diff
@@ -22,14 +22,14 @@ abstract class HealthCheck(private val webClient: WebClient) : HealthIndicator {
   }
 }
 
-@Component
+@Component("nomisApi")
```
andymarke, line 25: just a rename so all API have a constant naming convention

mikehalmamoj (approved):

PR #152, opened 2023-01-25
SDI-538 added API calls need for next part

andymarke:
* Adds all the endpoints required for create synchronisation 
* All non-blocking so we can see how that pans out


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingAdjustmentsApiService.kt
```diff
@@ -0,0 +1,28 @@
...
+  suspend fun getAdjustment(adjustmentId: String): AdjustmentDetails {
+    return sentenceAdjustmentsApiWebClient.get()
+      .uri("/adjustments/$adjustmentId")
+      .retrieve()
+      .bodyToMono(AdjustmentDetails::class.java)
+      .awaitSingle()
+  }
+}
+
+data class AdjustmentDetails(
```
andymarke, line 19: just a guess, API does not exist yet

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/helpers/SpringAPIServiceTest.kt
```diff
@@ -38,7 +38,7 @@ import kotlin.annotation.AnnotationTarget.CLASS
   VisitsApiExtension::class,
   IncentivesApiExtension::class,
   ActivitiesApiExtension::class,
-  SentenceAdjustmentsApiExtension::class
+  SentencingAdjustmentsApiExtension::class
 )
```
andymarke, line 42: rename since it deals with both types of adjustments

petergphillips (approved):
LGTM

PR #155, opened 2023-01-27
SDI-538 Recoverable error condition

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingDomainEventListener.kt
```diff
@@ -25,29 +27,35 @@ class SentencingDomainEventListener(
   }
 
   @SqsListener("sentencing", factory = "hmppsQueueContainerFactoryProxy")
-  fun onPrisonerChange(message: String) {
+  fun onPrisonerChange(
+    message: String,
+    visibility: Visibility
```
andymarke, line 32: temporary until SQS library supports this

mikehalmamoj (approved):

PR #156, opened 2023-01-30
SDI-538 Unrecoverable error condition

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingDomainEventListener.kt
```diff
@@ -53,9 +53,17 @@ class SentencingDomainEventListener(
...
-        visibility.changeTo(1)
+        visibility.changeTo(0)
         throw it
       }
-    }.thenAccept { }
+    }
   }
 }
+
+private fun asCompletableFuture(
```
andymarke, line 63: small readability change for above

mikehalmamoj: nice

mikehalmamoj (approved):

PR #165, opened 2023-02-02
SDI-508 Handle activities.prisoner.deallocated event

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesService.kt
```diff
@@ -110,6 +111,39 @@ class ActivitiesService(
...
+              // endComment = allocation.?, // TODO could put something useful in here
+            )
+          )
+        } catch (e: Exception) {
+          telemetryClient.trackEvent("activity-deallocate-failed", telemetryMap)
+          throw e
+        }
+
+        val mapWithNomisId = telemetryMap
+          .plus(Pair("offenderProgramReferenceId", nomisResponse.offenderProgramReferenceId.toString()))
```
mikehalmamoj, line 140: Why re-assign to a new value given the map is mutable?

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesServiceTest.kt
```diff
@@ -246,6 +246,84 @@ internal class ActivitiesServiceTest {
...
+
+      activitiesService.deallocate(
+        AllocationDomainEvent(
+          eventType = "dummy",
+          scheduleId = activityScheduleId,
+          allocationId = allocationId,
+          version = "1.0",
+          description = "description",
+          occurredAt = LocalDateTime.now(),
+        )
```
mikehalmamoj, line 275: I guess it doesn't matter as the events haven't been defined by the producer yet, but this doesn't fit the domain event schema - scheduleId and allocationId should be in an `additionalInformation` section and there should be a detail URL that points back to the Activities API.

I'd be slightly worried that the Activities team just copy this structure when they get around to producing events.

mikehalmamoj (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesService.kt
steverendell: good point, have changed to just put to the existing map

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesServiceTest.kt
steverendell: Done

mikehalmamoj (approved):

PR #167, opened 2023-02-06
SDI-613 Refactor message structure

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesService.kt
```diff
@@ -174,14 +174,18 @@ class ActivitiesService(
...
+  val additionalInformation: ScheduleAdditionalInformation,
   val version: String,
   val description: String,
   val occurredAt: LocalDateTime,
 )
 
+data class ScheduleAdditionalInformation(
+  val activityScheduleId: Long,
+)
+
```
mikehalmamoj, line 188: The message structure should align with the [HMPPS Domain Event Schema](https://dsdmoj.atlassian.net/wiki/spaces/NDSS/pages/3732570166/ADR+-+Domain+Event+Schema) (and the same change has been made in Activities). And a rename.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesDomainEventListener.kt
```diff
@@ -33,7 +36,7 @@ class ActivitiesDomainEventListener(
     val sqsMessage: SQSMessage = objectMapper.readValue(message)
     when (sqsMessage.Type) {
       "Notification" -> {
-        val (eventType) = objectMapper.readValue<HMPPSDomainEvent>(sqsMessage.Message)
+        val (eventType) = objectMapper.readValue<HmppsDomainEvent>(sqsMessage.Message)
```
mikehalmamoj, line 39: This was previously pointing at an Incentives domain event which has a different shape. All we need here is the event type and additional information could be different on each message anyway.

PR #179, opened 2023-02-08
SDI-544 Fixed unnecessary inclusion of Incentives additional info in …

andymarke:
…every domain events


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/listeners/HMPPSDomainEvent.kt
```diff
@@ -1,9 +1,5 @@
 package uk.gov.justice.digital.hmpps.prisonertonomisupdate.listeners
 
-import uk.gov.justice.digital.hmpps.prisonertonomisupdate.incentives.IncentivesService
-
```
andymarke, line 4: :-)

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/listeners/HMPPSDomainEvent.kt
```diff
@@ -1,9 +1,5 @@
...
 data class HMPPSDomainEvent(
   val eventType: String,
```
andymarke, line 4: we only use the eventType in the context this is used

mikehalmamoj (approved):

PR #187, opened 2023-02-15
SDI-610 Create activity schduled instances

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/NomisApiService.kt
```diff
@@ -267,6 +268,13 @@ data class CreateSentencingAdjustmentRequest(
   val comment: String?,
 )
 
+data class ScheduleRequest(
+  @JsonFormat(pattern = "yyyy-MM-dd")
+  val date: LocalDate,
+  val startTime: String,
+  val endTime: String,
```
steverendell, line 275: COuld use LocalTime?

lbennett-moj (approved):

PR #190, opened 2023-02-16
SDI-505 Fix Activities generated date format

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesApiServiceTest.kt
```diff
@@ -75,9 +75,9 @@ internal class ActivitiesApiServiceTest {
             },
             "startDate": "2022-12-30",
             "endDate": "2022-12-30",
-            "allocatedTime": "2022-12-30T14:03:06.365Z",
+            "allocatedTime": "2022-12-30T14:03:06",
```
mikehalmamoj, line 78: Activities are sending in this format which dolesn't parse to OffsetDateTime - so we're now using LocalDateTime instead (which is the same they  use on their side)

steverendell (approved):
lgtm

PR #195, opened 2023-02-22
SDI-607 Nomis screen only handles 12 long activity code

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesService.kt
```diff
@@ -214,9 +214,9 @@ class ActivitiesService(
 
   private fun toNomisActivityCode(activityDescription: String, scheduleDescription: String): String {
     var code = "SAA-$activityDescription"
-      .let { it.substring(0, min(it.length, 20)).trim() }
-    if (code.length < 19) {
-      code += "-${scheduleDescription.let { it.substring(0, min(it.length, 19 - code.length)) }}"
+      .let { it.substring(0, min(it.length, 12)).trim() }
```
andymarke, line 217: can we get duplicates and if we do what do we do about it?

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesService.kt
mikehalmamoj: Hmmm there's no constraint so I thought it wouldn't matter. However the UI has validation to prevent duplicates so I guess I need to emulate that 🤔

PR #205, opened 2023-03-03
SDIT-642 Use generic queue sender

andymarke:
First step to make mapping retry generic


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingRetryQueueService.kt
```diff
@@ -0,0 +1,20 @@
...
+import uk.gov.justice.hmpps.sqs.HmppsQueueService
+
+@Service
+class SentencingRetryQueueService(
+  hmppsQueueService: HmppsQueueService,
+  telemetryClient: TelemetryClient,
+  objectMapper: ObjectMapper,
+) :
+  RetryQueueService(
+    queueId = "sentencing",
```
mikehalmamoj, line 16: Should the queueId be injected from the properties config?

mikehalmamoj (approved):

PR #212, opened 2023-03-07
SDIT-642 DSL for synchronisation

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/SynchronisationService.kt
```diff
@@ -49,3 +50,83 @@ data class MappingTelemetry(
...
+    fetchMapping()?.let {
+      telemetryClient?.trackEvent(
+        "$name-create-duplicate",
+        eventTelemetry,
+        null,
+      )
+    } ?: let {
+      transform().also { mapping ->
+        mapping?.run {
+          val telemetryAttributes = eventTelemetry + mapping.asMap()
```
andymarke, line 73: almost certainly adding all mapping fields to telemetry is what we want

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/SynchronisationService.kt
```diff
@@ -3,11 +3,12 @@ package uk.gov.justice.digital.hmpps.prisonertonomisupdate.services
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.microsoft.applicationinsights.TelemetryClient
+import kotlin.reflect.full.memberProperties
 
 abstract class SynchronisationService(
```
mikehalmamoj, line 8: Shame this is abstract then you could let Spring wire up the dependencies. Can you make retryCreateMapping accept a lambda?

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingAdjustmentsService.kt
```diff
@@ -24,27 +24,30 @@ class SentencingAdjustmentsService(
...
+        "offenderNo" to createEvent.additionalInformation.nomsNumber,
+      )
+
+      checkMappingDoesNotExist {
+        sentencingAdjustmentsMappingService.getMappingGivenAdjustmentId(createEvent.additionalInformation.id)
+      }
+      transform {
+        val adjustment = sentencingAdjustmentsApiService.getAdjustment(createEvent.additionalInformation.id)
+
+        proceedWhen(adjustment.creatingSystem != CreatingSystem.NOMIS) {
```
mikehalmamoj, line 42: I'm not convinced about the need for this - is it possible for the entire transform function to return a nullable instead? null means we're not proceeding...

Also it might be easier to read if the entire transform parameter was a function call, e.g. `fun transformAdjustment(adjustmentId): SentenceAdjustmentMappingDto?`

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/SynchronisationService.kt
mikehalmamoj: I'm not sure about this - we pass a lot of data across in the Activities service and most of it is not telemetry. But I'm finding it hard to visualise.

andymarke: This is just the mapping object not the Activity - so just IDs

mikehalmamoj:
I think it's a promising start but needs trying on some other synchronisations. I think moving the transform lambda into a function will help readability and would like to see that tried.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/SynchronisationService.kt
andymarke: Actually I don't think I need this anymore. The `tryCreateMapping` is no longer used since the helper class does all that. So all it needs is an interface which implements `suspend fun retryCreateMapping(message: String)`

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingAdjustmentsService.kt
```diff
@@ -1,92 +1,74 @@
...
-            ),
-            null,
-          )
-        }
       }
+      saveMapping { sentencingAdjustmentsMappingService.createMapping(it) }
     }
   }
 
+  private suspend fun createTransformedAdjustment(adjustment: AdjustmentDetails) =
```
andymarke, line 50: reads better when this is extracted

mikehalmamoj (approved):

PR #229, opened 2023-03-22
SDI-639 Create a 'create-appointment' handler

steverendell:
Use real AppointmentInstance Dto now in activities api


lbennett-moj (approved):
LGTM

PR #248, opened 2023-04-12
SDIT-688 handle locked attendance status from Activities

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/AttendanceService.kt
```diff
@@ -30,10 +31,16 @@ class AttendanceService(
       val nomisCourseActivityId = mappingService.getMappingGivenActivityScheduleId(attendanceSync.activityScheduleId).nomisCourseActivityId
         .also { telemetryMap["nomisCourseActivityId"] = it.toString() }
 
+      val nomisAttendanceStatus = if (attendanceSync.status == "LOCKED") {
+        getNomisAttendanceStatus(nomisCourseActivityId, attendanceSync)?.eventStatus
+      } else {
+        null
+      }
```
mikehalmamoj, line 38: We only need the current Nomis attendance status if the Activities status is locked. Could have pushed this down into the method where `nomisAttendanceStatus` is used but that would involve passing various parameters down the call chain which just clutter things up - so chose to do it up front.

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/AttendanceServiceTest.kt
```diff
@@ -394,25 +446,28 @@ class AttendanceServiceTest {
     @ParameterizedTest
     @CsvSource(
       value = [
-        "WAITING,null,SCH",
-        "COMPLETED,CANCELLED,CANC",
-        "COMPLETED,null,COMP",
-        "LOCKED,null,EXP",
```
mikehalmamoj, line 400: Turns out these weren't being set to `null` but `"null"` instead 😱

steverendell (approved):
lgtm

PR #255, opened 2023-04-14
SDIT-739 added dummy cron to call reports endpoint

andymarke:


helm_deploy/hmpps-prisoner-to-nomis-update/templates/reports-cronjob.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ include "app.fullname" . }}-reports
+  labels:
+    {{- include "app.labels" . | nindent 4 }}
+spec:
+  schedule: "0 * * * *"
```
andymarke, line 8: every hour, but will change to every nighty once ready

helm_deploy/hmpps-prisoner-to-nomis-update/values.yaml
```diff
@@ -20,6 +20,10 @@ generic-service:
           deny all;
           return 401;
         }
```
andymarke, line 22: not sure if I need a `;` 😣

mikehalmamoj (approved):

PR #257, opened 2023-04-17
SDIT-743 updates and refactor

mikehalmamoj:
Nothing functional has changed. I've split the handlers for different message types into separate services and the same for the unit and integration tests. And tidied up some names in tests for consistency.


petergphillips (approved):
LGTM

PR #259, opened 2023-04-18
SDIT-740 Add edit/cancel/delete to-nomis event handlers

steverendell:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsToNomisIntTest.kt
```diff
@@ -453,8 +457,344 @@ class AppointmentsToNomisIntTest : SqsIntegrationTestBase() {
...
+          )
+        }
+      }
+    }
+
+    @Nested
+    inner class Exceptions {
+
+      @Nested
+      inner class WhenServiceFailsOnce {
```
mikehalmamoj, line 625: I'm not sure of the benefit of the fails once / fails forever tests on every single message - aren't we just testing that DLQ retries work? What do you think @andymarke?

mikehalmamoj (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsToNomisIntTest.kt
steverendell: I did wonder

andymarke: The original was there to test a specific scenario with SQS, also agreed we don't need it all over the place

PR #269, opened 2023-04-21
SDIT-739 Use own scope

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesResource.kt
```diff
@@ -33,15 +32,10 @@ class IncentivesResource(
     telemetryClient.trackEvent("incentives-reports-reconciliation-requested", mapOf("active-prisoners" to activePrisonersCount.toString()))
     log.info("Incentives reconciliation report requested for $activePrisonersCount active prisoners")
 
-    GlobalScope.launch {
-      runCatching {
```
andymarke, line 37: no longer need the runCatching since nothing in `incentivesReconciliationService.generateReconciliationReport` will every reasonably thrown anything

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/CoroutineScopes.kt
```diff
@@ -0,0 +1,12 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+class CoroutineScopes {
+  @Bean
+  fun reportScope() = CoroutineScope(Dispatchers.Unconfined)
+}
```
andymarke, line 12: everything we do is non-blocking so we don't care about sticking to a particular thread

mikehalmamoj (approved):

PR #277, opened 2023-04-25
Sdi 720 create mapping server endpoints to clear activities and appointments dev data

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/ResourceServerConfiguration.kt
```diff
@@ -27,6 +27,7 @@ class ResourceServerConfiguration {
           "/v3/api-docs/**", "/swagger-ui/**", "/swagger-ui.html",
           "/queue-admin/retry-all-dlqs",
           "/incentives/reports/reconciliation",
+          "/appointments", "/activities",
```
mikehalmamoj, line 30: You've got a role on the endpoints so we don't need to bypass security here? I don't think we're going to have a cronjob to perform the deletes are we?

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsResource.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.ResponseStatus
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+class AppointmentsResource(private val appointmentsService: AppointmentsService) {
+
+  /**
+   * For dev environment only - delete all appointments, for when activities environment is reset
+   */
```
mikehalmamoj, line 15: Should we feature switch to allow in dev only?

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsToNomisIntTest.kt
```diff
@@ -794,6 +794,38 @@ class AppointmentsToNomisIntTest : SqsIntegrationTestBase() {
     }
   }
 
+  @Nested
+  inner class DeleteAll {
```
mikehalmamoj, line 798: Not sure we should use this test class - it extends SqsIntegrationTestBase but this test is resource based instead of message based.

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/ResourceServerConfiguration.kt
steverendell: whoops good spot! will remove !!

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsToNomisIntTest.kt
steverendell: good idea

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsResource.kt
steverendell: definitely - will add

mikehalmamoj (approved):

PR #280, opened 2023-04-25
SDIT-422 Standardise telemetry

mikehalmamoj:
Standardise telemetry keys coming from DPS (activities) / Nomis. Also should use the same keys as the [API](https://github.com/ministryofjustice/hmpps-nomis-prisoner-api/pull/277).


steverendell (approved):
lgtm

PR #284, opened 2023-05-02
SDIT-412 Additional Amend Activity API details

mikehalmamoj:
Plus switch to Open API generated clients


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesService.kt
```diff
@@ -209,3 +214,11 @@ data class ScheduleDomainEvent(
...
   val activityScheduleId: Long,
 )
+
+fun ActivitySchedule.toCreatePayPerSession(): CreateActivityRequest.PayPerSession =
+  CreateActivityRequest.PayPerSession.values()
+    .first { it.value == this.activity.payPerSession.value }
+
+fun ActivitySchedule.toUpdatePayPerSession(): UpdateActivityRequest.PayPerSession =
+  UpdateActivityRequest.PayPerSession.values()
+    .first { it.value == this.activity.payPerSession.value }
```
mikehalmamoj, line 224: Despite reusing the same enum for PayPerSession the Open API generator creates multiple copies of it, so we need to map it the Activities enum for each copy.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/WebClientConfiguration.kt
```diff
@@ -28,12 +28,13 @@ class WebClientConfiguration(
   fun oauthApiHealthWebClient(): WebClient = WebClient.builder().baseUrl(oauthApiBaseUri).build()
 
   @Bean
-  fun nomisApiWebClient(authorizedClientManager: ReactiveOAuth2AuthorizedClientManager): WebClient {
+  fun nomisApiWebClient(authorizedClientManager: ReactiveOAuth2AuthorizedClientManager, webClientBuilder: WebClient.Builder): WebClient {
```
mikehalmamoj, line 31: We need to inject the WebClient.Builder that has been configured by Spring with the correct Codecs (specfically for date handling). We previously worked around this by using `JsonFormat(pattern = "yyyy-MM-dd")` on the request object but we can't do that now the request object is generated.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/NomisApiService.kt
```diff
@@ -343,113 +355,6 @@ data class CreateIncentiveResponseDto(
   val sequence: Long,
 )
 
-data class CreateActivityRequest(
```
mikehalmamoj, line 346: All the request objects are now generated by openapi-generator.

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/helpers/SpringAPIServiceTest.kt
```diff
@@ -43,6 +45,6 @@ import kotlin.annotation.AnnotationTarget.CLASS
   SentencingAdjustmentsApiExtension::class,
 )
 @ActiveProfiles("test")
-@SpringBootTest(classes = [WebClientConfiguration::class, WebClientAutoConfiguration::class, ReactiveOAuth2ClientAutoConfiguration::class, ReactiveSecurityAutoConfiguration::class])
+@SpringBootTest(classes = [JacksonAutoConfiguration::class, CodecsAutoConfiguration::class, WebClientConfiguration::class, WebClientAutoConfiguration::class, ReactiveOAuth2ClientAutoConfiguration::class, ReactiveSecurityAutoConfiguration::class])
```
mikehalmamoj, line 48: Some magic to make sure Spring wires up the WebClient.Builder correctly for these tests.

README.md
```diff
@@ -58,6 +58,29 @@ e.g.
...
+#### Updating the Open API specs
+
+Run the following commands to take a copy of the latest specs:
+
+```
+curl https://nomis-prsner-dev.aks-dev-1.studio-hosting.service.justice.gov.uk/v3/api-docs > openapi-specs/nomis-sync-api-docs.json
+curl https://activities-api-dev.prison.service.justice.gov.uk/v3/api-docs > openapi-specs/activities-api-docs.json
+```
+
+Go into the specs and reformat so they and the diffs are easier for humans to read.
```
petergphillips, line 74: sounds like hard work to reformat.  I would suggest either
1. Pipe the result through `jq .`
2. Use `http` instead with `--pretty=format`

petergphillips (approved):
LGTM

README.md
mikehalmamoj: Good shout - next time I update the specs I'll try to improve the process in this way.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/WebClientConfiguration.kt
```diff
@@ -28,12 +28,13 @@ class WebClientConfiguration(
...
     val oauth2Client = ServerOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager).also {
       it.setDefaultClientRegistrationId("nomis-api")
     }
 
-    return WebClient.builder()
+    return webClientBuilder
+      .clone() // clone so that we don't dirty the singleton builder
```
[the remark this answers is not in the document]
mikehalmamoj: Ah that's interesting. I did this based on a bug we had a couple of years ago where the builder was adding OAuth headers for health clients despite them not asking for one. But totally agree that shouldn't happen with a prototype bean. I'll investigate further in another PR but probably remove the clone.

PR #285, opened 2023-05-03
SDIT-412 improve openapi generator usage instructions

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/WebClientConfiguration.kt
```diff
@@ -34,7 +34,6 @@ class WebClientConfiguration(
     }
 
     return webClientBuilder
-      .clone() // clone so that we don't dirty the singleton builder
```
mikehalmamoj, line 37: If injecting the WebClient.Builder into the class then clone would make sense as it is indeed a singleton - injecting it into the method means we make use of the prototype bean so no need to clone.

petergphillips (approved):
LGTM

PR #289, opened 2023-05-12
SDIT-438 Handle suspension details

mikehalmamoj:


build.gradle.kts
```diff
@@ -75,7 +75,8 @@ tasks.register("buildActivityApiModel", GenerateTask::class) {
   configOptions.set(
     mapOf(
       "dateLibrary" to "java8-localdatetime",
-      "serializationLibrary" to "jackson"
+      "serializationLibrary" to "jackson",
+      "enumPropertyNaming" to "original"
```
mikehalmamoj, line 79: Workaround this [openapi generator bug](https://github.com/OpenAPITools/openapi-generator/issues/3804). Overrides the default which is to generate enums as cAMELCASE (but not camelCase as probably intended) 😆

lbennett-moj (approved):
LGTM

PR #300, opened 2023-06-02
SDIT-845 Use the Spring web client builder 

mikehalmamoj:
...to make use of property `spring.codec.max-in-memory-size` and prevent a `DataBufferLimitException`


andymarke (approved):
LGTM

PR #315, opened 2023-06-19

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsService.kt
```diff
@@ -153,8 +154,20 @@ class AppointmentsService(
...
 
+  private fun constructComment(instance: AppointmentInstance) =
+    if (instance.appointmentDescription == null) {
+      instance.comment
+    } else {
+      if (instance.comment == null) {
+        instance.appointmentDescription
+      } else {
+        "${instance.appointmentDescription} - ${instance.comment}"
+      }
```
lbennett-moj, line 168: Could use something like this to be idiomatic (don't know if it is any clearer!)
instance.appointmentDescription?.let{instance.comment?.let{ "${instance.appointmentDescription} - ${instance.comment}" } ?: instance.appointmentDescription} ?: instance.comment

lbennett-moj (approved):

PR #317, opened 2023-06-20

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsService.kt
```diff
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.prisonertonomisupdate.appointments
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.microsoft.applicationinsights.TelemetryClient
+import org.apache.commons.lang3.StringUtils.isBlank
```
mikehalmamoj, line 6: Kotlin has its own version of this - `isNullOrBlank` meaning that a 3rd party lib isn't required.

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsServiceTest.kt
```diff
@@ -54,15 +54,15 @@ internal class AppointmentsServiceTest {
 
   @Test
   fun `Comments are constructed correctly 3`() = runTest {
-    callService(null, "Comment")
+    callService("  ", "Comment")
```
mikehalmamoj, line 57: Thought you might keep the null test as well as blanks.

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsService.kt
steverendell: Ah thanks - missed that one. Will use it in next commit

PR #321, opened 2023-06-22

steverendell:
Truncate comment if too long


mikehalmamoj (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsService.kt
```diff
@@ -189,6 +189,8 @@ class AppointmentsService(
         "${instance.appointmentDescription} - ${instance.comment}"
       }
     }
+    return comment?.run { if (length <= 4000) this else substring(0, 4000) }
```
andymarke, line 192: very minor but `let` is probably the correct semantics, but either way I think you can do 

`return comment?.take(4000)`

(I thing that is the function name)

andymarke (dismissed):
lgtm but see comment

andymarke (approved):
lgtm

PR #326, opened 2023-06-26
SDIT-906 Handle partial scheduled instance response from Activities

mikehalmamoj:
Activities will soon return only recent and future scheduled instances instead of the full history which will become huge over time. This change maintains the Nomis/Activities schedule mappings in that scenario.


petergphillips (approved):
LGTM

PR #329, opened 2023-07-11
SDIT-939 Fix appointments create mapping retries

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsToNomisIntTest.kt
```diff
@@ -375,7 +378,9 @@ class AppointmentsToNomisIntTest : SqsIntegrationTestBase() {
           await untilAsserted {
             mappingServer.verify(
               2,
-              postRequestedFor(urlEqualTo("/mapping/appointments")),
+              postRequestedFor(urlEqualTo("/mapping/appointments"))
+                .withRequestBody(matchingJsonPath("appointmentInstanceId", equalTo("$APPOINTMENT_INSTANCE_ID")))
+                .withRequestBody(matchingJsonPath("nomisEventId", equalTo("$EVENT_ID"))),
```
mikehalmamoj, line 383: This test should have caught the bug but it wasn't checking the JSON to make sure the right details are being passed in the retry message.

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsToNomisIntTest.kt
```diff
@@ -356,6 +356,9 @@ class AppointmentsToNomisIntTest : SqsIntegrationTestBase() {
           appointmentsApi.stubGetAppointmentInstance(id = APPOINTMENT_INSTANCE_ID, response = appointmentResponse)
 
           publishAppointmentEvent("appointments.appointment-instance.created")
+
+          await untilCallTo { appointmentsApi.getCountFor("/appointment-instances/$APPOINTMENT_INSTANCE_ID") } matches { it == 1 }
+          await untilCallTo { nomisApi.postCountFor("/appointments") } matches { it == 1 }
```
mikehalmamoj, line 361: This is needed to make the test work - something weird happens with Wiremock when trying to verify too early and you can't assert on the JSON.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/DomainEventListener.kt
```diff
@@ -41,7 +43,15 @@ abstract class DomainEventListener(
...
-        RETRY_CREATE_MAPPING -> service.retryCreateMapping(sqsMessage.Message)
+        RETRY_CREATE_MAPPING -> runCatching { service.retryCreateMapping(sqsMessage.Message) }
+          .onFailure {
+            telemetryClient.trackEvent(
+              "create-mapping-retry-failure",
+              mapOf("retryMessage" to sqsMessage.Message),
+              null,
+            )
+            throw it
+          }
```
mikehalmamoj, line 54: This is to make it easier to find the mapping create retry failures - previously there was no telemetry which is confusing as there is telemetry for everything else that happens (or fails) in this service.

petergphillips (approved):
LGTM

PR #339, opened 2023-07-27
SDIT-969 Call back to adjudications service

andymarke:


build.gradle.kts
```diff
@@ -86,6 +86,27 @@ tasks.register("buildActivityApiModel", GenerateTask::class) {
   )
 }
 
+tasks.register("buildAdjudicationApiModel", GenerateTask::class) {
+  generatorName.set("kotlin")
+  skipValidateSpec.set(true)
```
andymarke, line 91: there swagger has some endpoints that we don't call that are invalid - so ignore them

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjustments/AdjudicationsService.kt
```diff
@@ -1,10 +1,50 @@
...
+      "prisonId" to prisonId,
+    )
+    synchronise {
+      name = "adjudication"
+      telemetryClient = this@AdjudicationsService.telemetryClient
+      retryQueueService = adjudicationRetryQueueService
+      eventTelemetry = telemetryMap
+
+      checkMappingDoesNotExist {
+        null
```
andymarke, line 28: no mapping integration yet, so pretend we don't have one

mikehalmamoj (approved):

PR #342, opened 2023-07-31
SDIT-969 Corrected incident dates and parse charge offence code

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjustments/AdjudicationsService.kt
```diff
@@ -70,8 +70,7 @@ internal fun ReportedAdjudicationResponseV2.toNomisAdjudication() = CreateAdjudi
     repairs = reportedAdjudication.damages.map { it.toNomisRepair() },
   ),
   charges = listOf(
-    // TODO convert charges
-    ChargeToCreate(offenceCode = "51:25C", offenceId = "${reportedAdjudication.adjudicationNumber + 1_000_000}/1"),
+    ChargeToCreate(offenceCode = reportedAdjudication.offenceDetails.offenceRule.nomisCode!!, offenceId = "${reportedAdjudication.adjudicationNumber + 1_000_000}/1"),
```
andymarke, line 73: should never be null when create event is raised so happy to allow NPE

petergphillips (approved):
LGTM

PR #344, opened 2023-07-31
SDIT-969 Use correct offence code when inciting another prisoner

andymarke:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationTransformationTest.kt
```diff
@@ -102,6 +102,64 @@ class AdjudicationTransformationTest {
...
+    )
+    val nomisAdjudication = dpsAdjudication.toNomisAdjudication()
+
+    assertThat(nomisAdjudication.charges).hasSize(1)
+    assertThat(nomisAdjudication.charges[0].offenceCode).isEqualTo("51:25D")
+    assertThat(nomisAdjudication.charges[0].offenceId).isEqualTo("2234567/1")
+  }
+
+  @Test
+  fun `assume there is a withOther offence code if this adjudication is the other prisoner in an incident`() {
```
mikehalmamoj, line 137: I was wondering about the !!s above... why would it be nullable in the API response but never in real life?

andymarke: It is the Adjudication DTO that can be nullable (this is generated from swagger )  when there is a mapping issue why creating an adjudication, but by the time it is created and we get event it should be OK

PR #360, opened 2023-09-01
Sdit 995 add non association create sync service to nomis

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/nonassociations/NonAssociationsService.kt
```diff
@@ -0,0 +1,98 @@
...
+  private val objectMapper: ObjectMapper,
+) : CreateMappingRetryable {
+
+  suspend fun createNonAssociation(event: NonAssociationDomainEvent) {
+    synchronise {
+      name = "nonAssociation"
+      telemetryClient = this@NonAssociationsService.telemetryClient
+      retryQueueService = nonAssociationsUpdateQueueService
+      eventTelemetry = mapOf("nonAssociationId" to event.additionalInformation.id.toString())
+
```
andymarke, line 34: @steverendell Will need to check that the NA hasn't been created in NOMIS, given we have 2-way sync I guess we would have this scenario:

1. NA created in NOMIS
2. NA Synchronised to DPS
3. DPS raises a create event 
And in this situation we need to ignore the create (and BTW the mapping may not be created yet since he sync has only just called the DPS endpoint)

For Sentence Adjustments we asked them to add something to message to indicate it was created by NOMIS so we can ignore the domain event

andymarke (approved):
LGTM

PR #368, opened 2023-09-07
SDIT-1081 Close non-association

steverendell:


openapi-specs/nomis-sync-api-docs.json
```diff
@@ -620,7 +631,7 @@
         }
       }
     },
-    "/non-associations/offender/{offenderNo}/ns-offender/{nsOffenderNo}/close": {
+    "/non-associations/offender/{offenderNo}/ns-offender/{nsOffenderNo}/sequence/{typeSequence}/close": {
       "put": {
```
karenmillermoj, line 635: Would we need to pass thru the typeSequence?  We know which one is open.  Or is this for convenience?

karenmillermoj (approved):
lgtm

openapi-specs/nomis-sync-api-docs.json
steverendell: I thought that as we know the sequence I'd pass it rather than having logic to find it on the nomis api side

PR #378, opened 2023-09-19
SDIT-1096 Non-association Reconciliation process

steverendell:
Initial PR for report with these limitations:
- only compares 2 fields of the NA so far
- only loops through NAs from Nomis; does not find extra NAs only in the new service

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/nonassociations/NonAssociationsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/nonassociations/NonAssociationsResource.kt
@@ -1,18 +1,32 @@
 package uk.gov.justice.digital.hmpps.prisonertonomisupdate.nonassociations
 
+import com.microsoft.applicationinsights.TelemetryClient
 import io.swagger.v3.oas.annotations.Hidden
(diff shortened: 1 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/nonassociations/NonAssociationsResource.kt
```diff
@@ -28,4 +42,29 @@ class NonAssociationsResource(
...
+        .onSuccess { listOfLists ->
+          val results = listOfLists.flatten()
+          log.info("Non-associations reconciliation report completed with ${results.size} mismatches")
+          val map = mapOf("mismatch-count" to results.size.toString(), "success" to "true") +
+            results.associate { "${it.id}" to "nomis=${it.nomisNonAssociation}, dps=${it.dpsNonAssociation}" }
+          telemetryClient.trackEvent("non-associations-reports-reconciliation-report", map)
+          log.info("Non-associations reconciliation report logged")
+        }
+        .onFailure {
+          telemetryClient.trackEvent("non-associations-reports-reconciliation-report", mapOf("success" to "false"))
```
lbennett-moj, line 65: Is this changing the convention of having success or failure in the event name?

PR #384, opened 2023-09-27
SDIT-1096 Non-association Reconciliation process

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/nonassociations/NonAssociationsReconciliationService.kt
```diff
@@ -28,12 +28,16 @@ class NonAssociationsReconciliationService(
   @Value("\${reports.non-associations.reconciliation.page-size}")
   private val pageSize: Long,
 ) {
+
+  var nomisTotalDetails = 0
```
steverendell, line 32: Need to correctly count all the na_details within each nomis NA

src/main/resources/application.yml
```diff
@@ -115,4 +115,4 @@ reports:
       page-size: 20
   non-associations:
     reconciliation:
-      page-size: 100
+      page-size: 20
```
steverendell, line 118: Slow but may be more reliable

andymarke (approved):
LGTM

PR #389, opened 2023-10-02
SDIT-1118 process for adjudication.evidence.updated event

andymarke:
TODO - actual implementation


helm_deploy/hmpps-prisoner-to-nomis-update/values.yaml
```diff
@@ -71,6 +71,12 @@ generic-service:
     FEATURE_EVENT_DELETEALL: false
     FEATURE_EVENT_ADJUDICATION_REPORT_CREATED: false
     FEATURE_EVENT_ADJUDICATION_DAMAGES_UPDATED: false
```
andymarke, line 73: just made this all consistent across ADJUDICATIONS - although everything is double typed an over the top with both defaults and entries for each environment. I suspect we could just drop the prod defaults but at least this is clear what to do when we go live

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsDomainEventListener.kt
```diff
@@ -37,6 +37,7 @@ class AdjudicationsDomainEventListener(
     when (eventType) {
       "adjudication.report.created" -> adjudicationsService.createAdjudication(message.fromJson())
       "adjudication.damages.updated" -> adjudicationsService.updateAdjudicationDamages(message.fromJson())
+      "adjudication.evidence.updated" -> log.info("Received a message that hasn't been wired up yet: {}", eventType)
       "adjudication.hearing.created" -> adjudicationsService.createHearing(message.fromJson())
```
andymarke, line 41: the TODO is here

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -398,14 +398,23 @@ private fun ReportedDamageDto.Code.toNomisEnum(): RepairToUpdateOrAdd.TypeCode =
   ReportedDamageDto.Code.CLEANING -> RepairToUpdateOrAdd.TypeCode.CLEA
   ReportedDamageDto.Code.REPLACE_AN_ITEM -> RepairToUpdateOrAdd.TypeCode.DECO // best match possibly?
 }
+private fun ReportedDamageDto.Code.toNomisCreateEnum(): RepairToCreate.TypeCode = when (this) {
```
andymarke, line 401: we have 2 enum list from NOMIS API that are identical - not sure the best way to handle this other than have duplicate code for the 2 return types

mikehalmamoj (approved):

PR #401, opened 2023-10-08
SDIT-1128 Upgrade localstack in SQS lib

steverendell:


andymarke (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/integration/IntegrationTestBase.kt
```diff
@@ -46,4 +48,14 @@ abstract class IntegrationTestBase {
     roles: List<String> = listOf(),
     scopes: List<String> = listOf(),
   ): (HttpHeaders) -> Unit = jwtAuthHelper.setAuthorisation(user, roles, scopes)
+
```
andymarke, line 51: Is this because we have some tests that are not SQS test but require localstack?

steverendell: Yes sort of, It's basically because the services have the retry service as a dependency, e.g. activitiesService requires activitiesRetryQueueService

PR #415, opened 2023-10-16
SDIT-1138 Create punishment implementation

andymarke:
TODO

1. Consecutive awards
2. Better understand and testing on mapping from DPS to NOMIS


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -364,10 +386,112 @@ class AdjudicationsService(
...
+      when (this.privilegeType) {
+        PunishmentDto.PrivilegeType.OTHER -> "Loss of ${this.otherPrivilege}"
+        else -> "Loss of ${this.privilegeType}"
+      }
+
+    DAMAGES_OWED -> "OTHER - Damages owed £${String.format("%.2f", this.damagesOwedAmount!!)}"
+    else -> null
+  }
+
+private fun PunishmentDto.toNomisSanctionStatus(): SanctionStatus {
```
andymarke, line 466: This is taken from prison-api but not sure it is correct - so need to validate and test better

lbennett-moj (approved):

PR #418, opened 2023-10-17
SDIT-1152 Ignore attendance paid error if paid today

mikehalmamoj:
If paid on the same day as the attendance (rather than by payroll overnight) then the payment was made by a process is outside the scope of DPS Activities, such as Discharge Balance Calculation. So we just ignore any updates rather than throw an error that cannot be fixed.


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/helpers/WebClientHelpers.kt
```diff
@@ -1,16 +1,34 @@
...
+        Mono.error(AttendancePaidException(errorResponse))
+      } else {
+        Mono.error(it)
+      }
+    }
+    .awaitSingle()
+
+enum class BadRequestError(val errorCode: Int) {
+  ATTENDANCE_PAID(1001),
+}
```
mikehalmamoj, line 30: I couldn't find a nice way to include the error enum in the Open API spec. so have duplicated here.

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/AttendancesIntTest.kt
```diff
@@ -131,9 +137,97 @@ class AttendancesIntTest : SqsIntegrationTestBase() {
...
+        verify(telemetryClient).trackEvent(
+          eq("activity-attendance-update-ignored"),
+          check<Map<String, String>> {
+            assertThat(it["reason"]).isEqualTo("Attendance update ignored as already paid session on ${LocalDate.now()} at 10:00")
+          },
+          isNull(),
+        )
+      }
+      assertThat(awsSqsActivityDlqClient.countMessagesOnQueue(activityDlqUrl).get()).isEqualTo(0)
+    }
```
mikehalmamoj, line 189: The important outcome here is that if attendance paid on same day it happened then we `ignore` the event.

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/AttendancesIntTest.kt
```diff
@@ -131,9 +137,97 @@ class AttendancesIntTest : SqsIntegrationTestBase() {
...
+      ).get()
+
+      await untilCallTo { ActivitiesApiExtension.activitiesApi.getCountFor("/synchronisation/attendance/$ATTENDANCE_ID") } matches { it == 1 }
+      await untilCallTo { NomisApiExtension.nomisApi.putCountFor("/schedules/$NOMIS_CRS_SCH_ID/booking/$NOMIS_BOOKING_ID/attendance") } matches { it == 1 }
+
+      // We now have a DLQ message and telemetry is raised
+      await untilAsserted {
+        verify(telemetryClient).trackEvent(eq("activity-attendance-update-failed"), any<Map<String, String>>(), isNull())
+      }
+      assertThat(awsSqsActivityDlqClient.countMessagesOnQueue(activityDlqUrl).get()).isEqualTo(1)
```
mikehalmamoj, line 229: The important outcome here is that if the attendance is updated after the day of attendance then we reject the update event and raise `failed` telemetry - meaning we don't understand why this update happened.

andymarke (approved):
LGTM

PR #419, opened 2023-10-18
SDIT-1138 Create punishment consecutive

andymarke:


openapi-specs/nomis-sync-api-docs.json
```diff
@@ -1,9510 +1 @@
...
-            "description": "The 2nd offender"
-          }
-        },
-        "description": "Non association id"
-      },
-      "PageNonAssociationIdResponse": {
-        "type": "object",
-        "properties": {
-          "totalPages": {
-            "type": "integer",
```
mikehalmamoj, line 1: Did you forget to push this through jq as per ther readme? Doesn't really matter I suppose but it keeps it human readable `curl https://nomis-prisoner-api-dev.prison.service.justice.gov.uk/v3/api-docs | jq . > openapi-specs/nomis-sync-api-docs.json`

andymarke: OK, point taken - which I will forget the next time I do this :-)

andymarke: we could have a "refresh-all" script to all at the same time

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -438,17 +439,19 @@ class AdjudicationsService(
...
-)
+  private suspend fun PunishmentDto.toNomisAward() = CreateHearingResultAwardRequest(
+    sanctionType = this.type.toNomisSanctionType(),
+    sanctionStatus = this.toNomisSanctionStatus(),
+    effectiveDate = this.schedule.startDate ?: this.schedule.suspendedUntil ?: LocalDate.now(),
+    sanctionDays = this.schedule.days,
+    compensationAmount = this.damagesOwedAmount?.toBigDecimal() ?: stoppagePercentage?.toBigDecimal(),
+    commentText = this.toComment(),
+    consecutiveCharge = this.consecutiveChargeNumber?.let { chargeNumber ->
+      adjudicationMappingService.getMappingGivenChargeNumber(chargeNumber).let { AdjudicationChargeId(it.adjudicationNumber, it.chargeSequence) }
```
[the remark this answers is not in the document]
andymarke: hmm, not sure - since the first it saying:
transform this if it exists
the second is saying:
Now I know i'm going to transform - get the data the convert

So I see them as nested - even though chaining would work perfectly, but you would have to change the `.let` to `?.let`

I think this is maybe just a personal style thing - and next week I'll do the opposite. 
Good point though

lbennett-moj (approved):

PR #427, opened 2023-10-23
SDIT-1154 handle deletion of referrals and referral outcomes

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -398,6 +409,46 @@ class AdjudicationsService(
...
+
+      val charge = adjudicationsApiService.getCharge(
+        eventData.chargeNumber,
+        eventData.prisonId,
+      )
+
+      val outcome = charge.reportedAdjudication.outcomes.firstOrNull()
+
+      // If no adjudication outcome exists - delete on NOMIS
+      // if outcome exists - update NOMIS with current state
```
lbennett-moj, line 431: the deletion of a referral **outcome** is a rollback to a referral state, in the NOMIS world. 
The deletion of a referral is treated like a deletion

PR #440, opened 2023-10-30
SDIT-1129: 🐛 Set response timeouts

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/WebClientConfiguration.kt
```diff
@@ -3,50 +3,55 @@ package uk.gov.justice.digital.hmpps.prisonertonomisupdate.config
...
+import reactor.netty.http.client.HttpClient
+import java.time.Duration
 
 @Configuration
 class WebClientConfiguration(
   @Value("\${api.base.url.nomis}") val nomisApiBaseUri: String,
   @Value("\${api.base.url.mapping}") val mappingBaseUri: String,
   @Value("\${api.base.url.oauth}") val oauthApiBaseUri: String,
+  @Value("\${api.health-timeout:2s}") val healthTimeout: Duration,
+  @Value("\${api.timeout:90s}") val timeout: Duration,
```
petergphillips, line 23: default to 90 seconds for non health requests

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/WebClientConfiguration.kt
```diff
@@ -57,9 +62,14 @@ class WebClientConfiguration(
...
   ): ReactiveOAuth2AuthorizedClientManager {
     val authorizedClientProvider = ReactiveOAuth2AuthorizedClientProviderBuilder.builder().clientCredentials().build()
-    val authorizedClientManager =
-      AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager(clientRegistrationRepository, oAuth2AuthorizedClientService)
-    authorizedClientManager.setAuthorizedClientProvider(authorizedClientProvider)
-    return authorizedClientManager
+    return AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager(
+      clientRegistrationRepository,
+      oAuth2AuthorizedClientService,
+    ).also { it.setAuthorizedClientProvider(authorizedClientProvider) }
```
mikehalmamoj, line 68: It technically makes no difference - but - if you're only working on the context object and there is no return value then I think [apply](https://kotlinlang.org/docs/scope-functions.html#apply) is intended for suitable for this purpose. (I always think of also indicating a side effect to "something else").

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/WebClientConfiguration.kt
petergphillips: yes good call, will amend in next PR

PR #441, opened 2023-10-30
SDIT-1129: 🎨 use apply instead of also

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/WebClientConfiguration.kt
```diff
@@ -46,8 +47,8 @@ class WebClientConfiguration(
     registrationId: String,
     url: String,
   ): WebClient {
-    val oauth2Client = ServerOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager).also {
-      it.setDefaultClientRegistrationId(registrationId)
+    val oauth2Client = ServerOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager).kotlinApply {
```
petergphillips, line 50: not as pretty as I was hoping, since `apply` is already a function on the filter function so had to introduce an import alias

PR #443, opened 2023-10-30
SDIT-1191 quash events subscriptions

andymarke:


src/main/resources/application-localstack.yml
```diff
@@ -35,12 +35,6 @@ hmpps.sqs:
       queueName: hmpps_prisoner_to_nomis_adjudication_queue
       dlqName: hmpps_prisoner_to_nomis_adjudication_dlq
       subscribeTopicId: domainevents
-      subscribeFilter: "{\"eventType\":[\"adjudication.report.created\", \"adjudication.damages.updated\", \"adjudication.evidence.updated\", \"adjudication.hearing.created\",\
```
andymarke, line 38: these give no value running locally other then a headache keeping them up to date

src/test/resources/application-test.yml
```diff
@@ -66,12 +66,6 @@ hmpps.sqs:
       queueName: "adjudication-${random.uuid}"
       dlqName: "adjudication-dlq-${random.uuid}"
       subscribeTopicId: domainevents
-      subscribeFilter: "{\"eventType\":[\"adjudication.report.created\", \"adjudication.damages.updated\", \"adjudication.evidence.updated\", \"adjudication.hearing.created\",\
```
andymarke, line 69: even less  no value when running tests other then a headache keeping them up to date

andymarke: We don't have any tests that fire an event not in that list and assert we never received it

mikehalmamoj (approved):

PR #447, opened 2023-10-31
SDIT-1136 Document Activities reconciliation failures

mikehalmamoj:


README.md
```diff
@@ -266,6 +266,81 @@ The side effect of this event is a DPS IEP Review is created which is then writt
...
+
+DPS:
+```sql
+select * from allocations 
+where booking_id=<insert offender book id here>
+and active=true;
+```
+
+To get an overview of what's been happening with the synchronisation events for this offender run the following query in App Insights:
+```ksql
```
petergphillips, line 296: hadn't realised that `ksql` was a type - I think you've used `kusto` previously that didn't work

petergphillips (approved):
LGTM

README.md
mikehalmamoj: I don't think any of them do anything tbh. It's just a clue for the reader.

PR #461, opened 2023-11-13
SDIT-1217 Fix reconciliation telemetry

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesReconService.kt
```diff
@@ -115,47 +115,57 @@ class ActivitiesReconService(
...
+      }
+
+    // Publish success telemetry if all failures were ignored or there were no failures
+    if (differences.all().size - ignoredBookings.size == 0) {
       telemetryClient.trackEvent(
         "activity-$type-reconciliation-report-success",
         mutableMapOf("prison" to prisonId).apply {
           date?.also { this["date"] = "$it" }
         },
       )
```
mikehalmamoj, line 159: This is the fix - previously if all failures were ignored then we didn't publiksh the success telemetry, but we should.

andymarke (approved):
lgtm

PR #467, opened 2023-11-20
SDIT-1256 No longer send adjudication number

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -1090,7 +1090,6 @@ private fun PunishmentsAdditionalInformation.toTelemetryMap(): MutableMap<String
 )
 
 internal fun ReportedAdjudicationResponse.toNomisAdjudication() = CreateAdjudicationRequest(
-  adjudicationNumber = reportedAdjudication.chargeNumber.toNomisAdjudicationNumber(),
```
andymarke, line 1093: no longer supply adjudicationNumber or the companion offenceId, these are now generated by nomis api and returned

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -102,7 +102,7 @@ class AdjudicationsService(
 
         AdjudicationMappingDto(
           adjudicationNumber = nomisAdjudicationResponse.adjudicationNumber,
-          chargeSequence = nomisAdjudicationResponse.adjudicationSequence,
+          chargeSequence = nomisAdjudicationResponse.charges.first().chargeSequence,
```
andymarke, line 105: unrelated, but this code was wrong but would have always worked since both values would always be "1" - but use the correct mapping anyway to avoid confusion (there is always one and only one charge)

lbennett-moj (approved):

PR #469, opened 2023-11-23
SDIT-1265 Add repair endpoint to synchronise current incentive level to NOMIS

andymarke:
Required in the very rare scenario when to Incentives are added in quick succession and their is a network issue

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesDataRepairResourceIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesDataRepairResourceIntTest.kt
@@ -1,0 +1,102 @@
+package uk.gov.justice.digital.hmpps.prisonertonomisupdate.incentives
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesDataRepairResourceIntTest.kt
```diff
@@ -0,0 +1,102 @@
...
+    inner class HappyPath {
+      @Test
+      fun `repair incentive`() {
+        webTestClient.post().uri("/incentives/prisoner/booking-id/{bookingId}/repair", bookingId)
+          .headers(setAuthorisation(roles = listOf("ROLE_NOMIS_INCENTIVES")))
+          .contentType(MediaType.APPLICATION_JSON)
+          .exchange()
+          .expectStatus().isOk
+
+        verify(telemetryClient, times(1)).trackEvent(eq("incentives-repair"), telemetryCaptor.capture(), isNull())
```
mikehalmamoj, line 91: No assertions that it actually called the Nomis API to create?

mikehalmamoj (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesDataRepairResourceIntTest.kt
andymarke: yeah, lazy - I relied on telemetry reporting to correct thing - I'll improve

PR #472, opened 2023-11-27
SDIT-1264 Use real NOMIS adjustment DTOs

andymarke:
* also add convenience extension for WireMock to supply objects for bodies


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/wiremock/NomisApiMockServer.kt
```diff
@@ -15,16 +17,20 @@ import org.junit.jupiter.api.extension.BeforeAllCallback
...
 class NomisApiExtension : BeforeAllCallback, AfterAllCallback, BeforeEachCallback {
   companion object {
     @JvmField
     val nomisApi = NomisApiMockServer()
+    lateinit var objectMapper: ObjectMapper
   }
 
   override fun beforeAll(context: ExtensionContext) {
+    objectMapper = (SpringExtension.getApplicationContext(context).getBean("jacksonObjectMapper") as ObjectMapper)
     nomisApi.start()
```
andymarke, line 34: for now, assume there is always a Spring Context when we use these extensions

mikehalmamoj (approved):

PR #473, opened 2023-11-27
SDIT-1275 Handle nullable allocation prison pay band

mikehalmamoj:


openapi-specs/activities-api-docs.json
```diff
@@ -7245,7 +7245,6 @@
           "bookingId",
           "id",
           "isUnemployment",
-          "prisonPayBand",
```
mikehalmamoj, line 7248: This change is coming soon and when it does it will break the existing API. Therefore I decided to get ahead of the curve by making it nullable now, meaning that I can release ahead of the breaking change so that it won't be breaking.

andymarke (approved):
lgtm

PR #479, opened 2023-11-30
SDIT-1264 Catch 422 adjustments errors 

andymarke:
422 error from DPS adjustments means prisoner has no active sentence so can't return adjustments. For now we will treat these as no adjustments and see how that pans out - though I guess we will just skip them eventually.


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesMappingService.kt
```diff
@@ -6,7 +6,7 @@ import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.awaitBodilessEntity
 import org.springframework.web.reactive.function.client.awaitBody
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.helpers.awaitBodilessEntityOrThrowOnConflict
-import uk.gov.justice.digital.hmpps.prisonertonomisupdate.helpers.awaitBodyOrNotFound
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.helpers.awaitBodyOrNullForNotFound
```
andymarke, line 9: just renamed this since I found it confusing.

lbennett-moj (approved):

PR #480, opened 2023-11-30
SDIT-1264 Filter DPS adjustments by booking until they fix

andymarke:
DPS is incorrectly returning old adjustments that are no longer relevant, they will fix this - but for now we just filter by bookingId


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingReconciliationService.kt
```diff
@@ -57,7 +57,11 @@ class SentencingReconciliationService(
   suspend fun checkBookingAdjustmentsMatch(prisonerId: ActivePrisonerId): MismatchSentencingAdjustments? = runCatching {
     val (nomisAdjustments, dpsAdjustments) = withContext(Dispatchers.Unconfined) {
       async { nomisApiService.getAdjustments(prisonerId.bookingId) } to
-        async { adjustmentsApiService.getAdjustments(prisonerId.offenderNo) }
+        async {
+          adjustmentsApiService.getAdjustments(prisonerId.offenderNo)
+            // temporary fix until DPS filter out old adjustments
+            ?.filter { it.bookingId == prisonerId.bookingId }
```
andymarke, line 63: main change here - medium term we shouldn't need to do this once DPS API is fixed

PR #497, opened 2024-01-15
SDIT-1408: 🎨 Upgrade to latest ktlint

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/NomisApiService.kt
```diff
@@ -751,9 +752,7 @@ data class NomisIncentive(
 
 data class NomisCodeDescription(val code: String, val description: String)
 
-class RestResponsePage<T>
-@JsonCreator(mode = JsonCreator.Mode.PROPERTIES)
-constructor(
+class RestResponsePage<T>(
```
petergphillips, line 755: removed as is the only constructor anyway

lbennett-moj (approved):
lgtm

PR #506, opened 2024-01-31
SDIT-1456 Reconciliation retry wrapper does not work

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/HelperApi.kt
```diff
@@ -9,5 +9,5 @@ suspend fun <T> doApiCallWithSingleRetry(api: suspend () -> T) = runCatching {
   api()
 }.onFailure {
   log.warn("Retrying API call", it)
-  api()
+  return api()
```
steverendell, line 12: Lack of a 'return' meant the exception was thrown anyway! 🙄

lbennett-moj (approved):

PR #511, opened 2024-02-02
SDIT-1469 Temporarily increase max response size

mikehalmamoj:
Until we put a proper fix in - call a better endpoint for `/activities/{activityId}` that doesn't return all the schedules.


andymarke (approved):
lgtm

PR #513, opened 2024-02-02
SDIT-1469 Preempt breaking change

mikehalmamoj:


openapi-specs/activities-api-docs.json
```diff
@@ -9417,13 +9416,6 @@
               "$ref": "#/components/schemas/ActivityEligibility"
             }
           },
-          "schedules": {
-            "type": "array",
-            "description": "A list of schedules for this activity. These contain the time slots / recurrence settings for instances of this activity.",
-            "items": {
-              "$ref": "#/components/schemas/ActivitySchedule"
-            }
-          },
```
mikehalmamoj, line 9426: Manually removing this before it gets removed from the Activities API meanse that the activities change won't be a breaking change for us, so we won't have to coordinate releases.

steverendell (approved):
lgtm

PR #515, opened 2024-02-05
SDIT-1264 Added retry to adjustment reconciliation and renamed helper

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingReconciliationService.kt
```diff
@@ -41,7 +42,7 @@ class SentencingReconciliationService(
   }
 
   private suspend fun getActivePrisonersForPage(page: Pair<Long, Long>) =
-    runCatching { nomisApiService.getActivePrisoners(page.first, page.second).content }
+    runCatching { doApiCallWithRetries { nomisApiService.getActivePrisoners(page.first, page.second) }.content }
```
andymarke, line 45: main change is here and the 2 calls below

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/HelperApi.kt
```diff
@@ -5,7 +5,7 @@ import org.slf4j.LoggerFactory
 
 val log: Logger = LoggerFactory.getLogger("HelperApi")
 
-suspend fun <T> doApiCallWithSingleRetry(api: suspend () -> T) = runCatching {
+suspend fun <T> doApiCallWithRetries(api: suspend () -> T) = runCatching {
```
mikehalmamoj, line 8: I know this works but isn't it something we should be doing at the web client level? https://www.baeldung.com/spring-webflux-retry#1-usingretry

PR #553, opened 2024-03-19
SDIT-1597 Handle `prisoner-alerts.alert-deleted` domain event

andymarke:
Will try to delete if delete originated from DPS and mapping still exists. 
Will pragmatically ignore delete mapping failures

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsService.kt
@@ -3,6 +3,8 @@ package uk.gov.justice.digital.hmpps.prisonertonomisupdate.alerts
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.microsoft.applicationinsights.TelemetryClient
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsService.kt
```diff
@@ -96,8 +102,40 @@ class AlertsService(
...
+      runCatching {
+        mappingApiService.getOrNullByDpsId(dpsAlertId)?.also { mapping ->
+          telemetryMap["nomisBookingId"] = mapping.nomisBookingId.toString()
+          telemetryMap["nomisAlertSequence"] = mapping.nomisAlertSequence.toString()
+
+          nomisApiService.deleteAlert(bookingId = mapping.nomisBookingId, alertSequence = mapping.nomisAlertSequence)
+          tryToDeletedMapping(dpsAlertId)
+          telemetryClient.trackEvent("alert-deleted-success", telemetryMap, null)
+        } ?: also {
+          telemetryClient.trackEvent("alert-deleted-skipped", telemetryMap, null)
```
petergphillips, line 123: trivial, but looks a bit odd - don't think `null` is needed here (see line 130)

petergphillips (approved):
LGTM

PR #555, opened 2024-03-20
SDIT-1497 To Nomis locations sync updates

steverendell:
Update to breaking change in locations api

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -134,3 +134,8 @@ reports:
   non-associations:
     reconciliation:
       page-size: 40
+
+hmpps.sqs:
+  queues:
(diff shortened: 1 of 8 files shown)


src/main/resources/application.yml
```diff
@@ -134,3 +134,8 @@ reports:
   non-associations:
     reconciliation:
       page-size: 40
+
+hmpps.sqs:
+  queues:
+    location:
+      errorVisibilityTimeout: 30
```
steverendell, line 141: Also adjust queue retries so it doesnt try again too quickly. These can fail due to a missing parent location

andymarke (approved):
LGTM

PR #562, opened 2024-04-02
SDIT-1497 To Nomis locations sync updates - date fix

steverendell:
Fix for occurredAt datetime with BST offset


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/locations/LocationsService.kt
```diff
@@ -411,7 +410,6 @@ data class LocationDomainEvent(
   val eventType: String,
   val version: String,
   val description: String,
-  val occurredAt: LocalDateTime,
```
steverendell, line 414: It fails to de-serialise the json date in the message into this field if there is a +01:00 on the end. Luckily we dont need it.

andymarke (approved):
Lgtm

PR #563, opened 2024-04-02
Remove unused 'occurredAt' message fields

steverendell:
They were declared as the wrong type anyway: the occurredAt field is an OffsetDateTime.


petergphillips (approved):
LGTM

PR #565, opened 2024-04-03
SDIT-1625 Ignore Adjudication mismatch for merges

andymarke:
If there is an ADA mismatch and there has been a merge since migration and there is an adjudication missing on the active DPS booking then ignore mismatch.

TODO - can make more sophisticated by finding the missing adjudication and find the actual number of ADA days


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsReconciliationService.kt
```diff
@@ -88,6 +92,43 @@ class AdjudicationsReconciliationService(
...
+          ),
+        )
+        return false
+      } else {
+        return true
+      }
+    } else {
+      return true
+    }
+  }
```
andymarke, line 126: I will clean  this if else ugliness up at some point....

petergphillips (approved):
LGTM

PR #567, opened 2024-04-12
SDIT-1619 Locations reconciliation

steverendell:
Add details of history discrepancies

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/locations/LocationsReconciliationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/locations/LocationsReconciliationService.kt
@@ -10,6 +10,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.config.trackEvent
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.locations.model.ChangeHistory
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.locations.model.Location
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomismappings.model.LocationMappingDto
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomissync.model.AmendmentResponse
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/locations/LocationsReconciliationService.kt
```diff
@@ -239,10 +240,89 @@ class LocationsReconciliationService(
...
+  fun historyDoesNotMatch(nomis: LocationResponse, dps: Location): Boolean {
+    if (expectedDpsHistorySize(nomis.amendments) == (dps.changeHistory?.size ?: 0)) {
+      return false
+    }
+    val nomisListConverted = nomis.amendments!!.map { a ->
+      ChangeHistory(toHistoryAttribute(a.columnName), a.amendedBy, a.amendDateTime, a.oldValue, a.newValue)
+    }
+    val nomisSetSorted = nomisListConverted.sortedBy { it.amendedDate + it.attribute + it.amendedBy + it.oldValue + it.newValue }
+    val dpsSetSorted = dps.changeHistory!!.sortedBy { it.amendedDate + it.attribute + it.amendedBy + it.oldValue + it.newValue }
+
```
andymarke, line 260: minor Kotlin style things

...
You have 2 list so you could use zip to combine them then you don't need the `dpsSetSorted.elementAtOrNull(dpsIndex)`

you could of also done `forEachIndexed`

and lastly you could add a `isSame` method on either of the NOMIS or DPS objects to  then you could use something like list.any or something similar
or alternatively a extension function that returns the same data class from both objects and then use `==`

Obviously code works now but you might want to use Kotlin to improve

andymarke (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/locations/LocationsReconciliationService.kt
steverendell: Actually this comparison code was generated by the Intellij AI ! And it did originally use forEachIndexed but I needed to manipulate the list position at one point. I left all of this in partly because I might need it in future (the locations req are still somewhat influx).
I like the idea of using zip though - will try it next time

andymarke (approved):
lgtm

PR #581, opened 2024-05-13
SDIT-1730 Ignore alerts that should be inactive in NOMIS

andymarke:
Fix issue where there are two active alerts but one of them has expired

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsReconciliationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsReconciliationService.kt
@@ -57,8 +57,8 @@ class AlertsReconciliationService(
   suspend fun checkActiveAlertsMatch(prisonerId: ActivePrisonerId): MismatchAlerts? = runCatching {
     val nomisActiveAlerts = doApiCallWithRetries { nomisAlertsApiService.getAlertsForReconciliation(prisonerId.offenderNo) }.let { it.latestBookingAlerts + it.previousBookingsAlerts }
     val nomisAlerts = nomisActiveAlerts.map { it.alertCode.code }.toSortedSet()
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsReconciliationService.kt
```diff
@@ -57,8 +57,8 @@ class AlertsReconciliationService(
   suspend fun checkActiveAlertsMatch(prisonerId: ActivePrisonerId): MismatchAlerts? = runCatching {
     val nomisActiveAlerts = doApiCallWithRetries { nomisAlertsApiService.getAlertsForReconciliation(prisonerId.offenderNo) }.let { it.latestBookingAlerts + it.previousBookingsAlerts }
     val nomisAlerts = nomisActiveAlerts.map { it.alertCode.code }.toSortedSet()
-    val expiredActiveNomisAlerts = nomisActiveAlerts.filter { it.shouldBeInactive() }.map { it.alertCode.code }.toSortedSet()
-    val nomisNotExpiredAlerts = nomisAlerts - expiredActiveNomisAlerts
+    val nomisNotExpiredAlerts = nomisActiveAlerts.filterNot { it.shouldBeInactive() }.map { it.alertCode.code }.toSortedSet()
+    val expiredActiveNomisAlerts = nomisAlerts - nomisNotExpiredAlerts
     val dpsAlerts = doApiCallWithRetries { dpsAlertsApiService.getActiveAlertsForPrisoner(prisonerId.offenderNo) }.map { it.alertCode.code }.toSortedSet()
```
andymarke, line 62: reverse "not" logic to maintain list of real active alerts before checking which ones shouldn't be checked.
See test case below which would fail with the old logic

mikehalmamoj: It's a bit double negative - `filterNot { it.shouldBeInactive() }`

mikehalmamoj (approved):

PR #585, opened 2024-05-20
SDIT-1762 Use Adjudication duration rather than days

andymarke:
`days` is deprecated. Also with the unit measurement is hours set duration to 0 since NOMIS does not understand hours


petergphillips (approved):
LGTM

PR #608, opened 2024-06-21
SDIT-1475 remove domain event reason

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsService.kt
```diff
@@ -155,19 +155,13 @@ data class AlertAdditionalInformation(
   val prisonNumber: String,
   val alertCode: String,
   val source: AlertSource,
-  val reason: AlertReason,
```
andymarke, line 158: has been removed by DPS

lbennett-moj (approved):

PR #615, opened 2024-07-04
SDIT-1510 wire up DPS event to create court case

lbennett-moj:


helm_deploy/hmpps-prisoner-to-nomis-update/values.yaml
```diff
@@ -94,7 +94,7 @@ generic-service:
     FEATURE_EVENT_INCENTIVES_LEVELS_REORDERED: true
     FEATURE_EVENT_INCENTIVES_PRISON_LEVEL_CHANGED: true
     FEATURE_EVENT_DELETEALL: false
-    FEATURE_EVENT_COURT_SENTENCING_COURT_CASE_CREATED: true
+    FEATURE_EVENT_COURT_CASE_INSERTED: true
```
andymarke, line 97: minor point; but I think this is redundant since the by default it will be true - so I don't add these anymore

andymarke (approved):
lgtm

PR #635, opened 2024-07-25
Remove unused field

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesApiService.kt
```diff
@@ -69,7 +57,6 @@ data class IepDetail(
   val iepDate: LocalDate,
   val iepTime: LocalDateTime,
   val agencyId: String,
-  val locationId: String? = null,
```
mikehalmamoj, line 72: This is the change. Not sure why ktlint is so keen to reformat everything these days but ignore everything above.

steverendell (approved):
lgtm

PR #637, opened 2024-07-29
SDIT-1854 Casenotes sync to nomis part 3

steverendell:
Casenotes service

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesService.kt
@@ -1,0 +1,179 @@
+package uk.gov.justice.digital.hmpps.prisonertonomisupdate.casenotes
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.microsoft.applicationinsights.TelemetryClient
+import org.slf4j.Logger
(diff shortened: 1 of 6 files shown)


andymarke (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesService.kt
```diff
@@ -0,0 +1,179 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.config.trackEvent
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomismappings.model.CaseNoteMappingDto
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.CreateMappingRetryMessage
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.CreateMappingRetryable
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.synchronise
+
+@Service
+class CaseNotesService(
```
mikehalmamoj, line 16: Seems like a lot of code going in without tests... I guess they're coming later? But makes it really hard to review both code and tests.

steverendell: Good point, yes I thought Id get this boilerplate stuff in now. And due to the uncertainty about what the real casenotes api and events will look like there could be significant changes

PR #638, opened 2024-07-31
SDIT-1917 Introduce Mono retry feature on some endpoints used by reconciliation

steverendell:
Cut load on locations-api and catch additional network exception

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/locations/LocationsReconciliationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/locations/LocationsReconciliationService.kt
@@ -151,7 +151,7 @@ class LocationsReconciliationService(
 
     val (nomisRecord, dpsRecord) = withContext(Dispatchers.Unconfined) {
       async { nomisApiService.getLocationDetails(mapping.nomisLocationId) } to
-        async { locationsApiService.getLocation(mapping.dpsLocationId, true) }
+        async { locationsApiService.getLocation(mapping.dpsLocationId, false) }
     }.awaitBoth()
 
     val verdict = doesNotMatch(nomisRecord, dpsRecord)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/locations/LocationsReconciliationService.kt
```diff
@@ -151,7 +151,7 @@ class LocationsReconciliationService(
 
     val (nomisRecord, dpsRecord) = withContext(Dispatchers.Unconfined) {
       async { nomisApiService.getLocationDetails(mapping.nomisLocationId) } to
-        async { locationsApiService.getLocation(mapping.dpsLocationId, true) }
+        async { locationsApiService.getLocation(mapping.dpsLocationId, false) }
```
steverendell, line 154: Turn off history retrieval - which we are no longer checking

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/RetryApiService.kt
```diff
@@ -32,7 +33,8 @@ class RetryApiService(
     // Timeout for NO_RESPONSE is wrapped in a WebClientRequestException
     it is ReadTimeoutException || it.cause is ReadTimeoutException ||
       it is ConnectTimeoutException || it.cause is ConnectTimeoutException ||
-      it is SocketException || it.cause is SocketException
+      it is SocketException || it.cause is SocketException ||
+      it is PrematureCloseException || it.cause is PrematureCloseException
```
steverendell, line 37: PrematureCloseException appears to be thrown when the network 'connection is closed' error occurs

src/main/resources/application.yml
```diff
@@ -148,7 +148,7 @@ reports:
       page-size: 40
   locations:
     reconciliation:
-      page-size: 40
+      page-size: 20
```
steverendell, line 151: It turns out most of the other reconciliations already use 20 🤭 
This should decrease the load on the locations-api

PR #644, opened 2024-08-07
SDIT-1465 Turn-on all alert domain events

andymarke:
The person alerts where all on my accident due to domain name change


petergphillips (approved):
LGTM

PR #681, opened 2024-09-02
SDIT-2008 Refactor to make room for lots of diferent prison person syncs

mikehalmamoj:
I've tried to achieve the following:
* remove noise from class names (e.g. remove PrisonPerson from all class names, it blinds me)
* pull physical attributes related stuff into its own package (so it's easier to extend for new fields)

No code has actually been deleted, just shuffled around and renamed.


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/prisonperson/ReconciliationService.kt
```diff
@@ -5,21 +5,24 @@ import kotlinx.coroutines.Dispatchers
...
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.NomisApiService as PrisonerNomisApiService
 
-@Service
-class PrisonPersonReconService(
-  private val dpsApi: PrisonPersonDpsApiService,
-  private val prisonPersonNomisApi: PrisonPersonNomisApiService,
-  private val nomisApi: NomisApiService,
+@Service("prisonPersonReconciliationService")
+class ReconciliationService(
+  @Qualifier("physicalAttributesDpsApiService") private val dpsApi: DpsApiService,
```
mikehalmamoj, line 23: This is an anomaly - we have to call the DPS physical attributes endpoint when doing the reconciliation. With this approach we'll end up calling a lot of services for all of the different fields.

I'll suggest that we have a dedicated reconciliation endpoint, the same as we do for getting the NOMIS reconciliation data.

petergphillips (approved):
LGTM

PR #691, opened 2024-09-06
SDIT-2030 Ignore appointment deletion sync events where appointment does not exist in Nomis

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsService.kt
```diff
@@ -111,15 +112,21 @@ class AppointmentsService(
...
+        mappingService.getMappingGivenAppointmentInstanceId(appointmentInstanceId).nomisEventId
           .also { telemetryMap["nomisEventId"] = it.toString() }
 
-      nomisApiService.deleteAppointment(nomisEventId)
+      try {
+        nomisApiService.deleteAppointment(nomisEventId)
+      } catch (e: WebClientResponseException.NotFound) {
+        telemetryClient.trackEvent("appointment-delete-missing-ignored", telemetryMap, null)
+      }
+      mappingService.deleteMapping(appointmentInstanceId)
```
steverendell, line 129: Also now deleting the mapping. I think I omitted this originally thinking deletions were rare and it might aid troubleshooting to keep the old mapping.

andymarke: I think we would expect nomisApiService.deleteAppointment to be idempotent and not a throw a 404  exception if not found and you are going to ignore.

I think a better approach might have been

1) Check if we have a mapping
2) If not appisnights log the ignore
3) If does delete

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/appointments/AppointmentsService.kt
steverendell: Confirmed that things are ok as they are for this scenario (as it is handling a problem caused by a missing 2 way sync similar to visit cancellations)

PR #694, opened 2024-09-07
SDIT-2029 Turn on prison person reconciliation job in dev and prod

mikehalmamoj:


helm_deploy/values-preprod.yaml
```diff
@@ -85,3 +85,4 @@ generic-prometheus-alerts:
 
 cron:
   activitiesTidyMappingsSuspend: false
+  prisonPersonReconsSuspend: true
```
mikehalmamoj, line 88: reconciliation suspended in preprod because a) we know it's out of sync prior to the weekend DB refresh and b) not sure there's any value in running the reconciliation in this env anyway.

brightonsbox (approved):
LGTM

PR #702, opened 2024-09-18
SDIT-1898 Switch mock casenotes-api to real one

steverendell:
Preliminary work just getting nomis-api up to date


andymarke (approved):
lgtm

PR #712, opened 2024-09-30
SDIT-1854 Casenotes sync to nomis - fix and int tests

steverendell:
Mainly int tests but also including:
- fix to pass new author username field to Nomis
- add alerts for casenotes queues
- correct field name in telemetry


petergphillips (approved):
LGTM

PR #725, opened 2024-10-03
SDIT-2093 Handle csip updated domain event

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/csip/CSIPService.kt
```diff
@@ -63,6 +64,41 @@ class CSIPService(
...
+            nomisCSIPReportId = nomisCsip.nomisCSIPReportId,
+            mappingType = CSIPFullMappingDto.MappingType.DPS_CREATED,
+            attendeeMappings = listOf(),
+            factorMappings = listOf(),
+            interviewMappings = listOf(),
+            planMappings = listOf(),
+            reviewMappings = listOf(),
+          )
+        }
+      // TODO Check and update child mappings
```
karenmillermoj, line 94: There is still additional work to do here - as new child mappings may have been created - will do in a separate PR

mikehalmamoj (approved):

PR #726, opened 2024-10-03
SDIT-2023 Synchronise height and weight from DPS when a prisoner is readmitted on an old booking

mikehalmamoj:
This handles an edge case where a prisoner is recalled but received on a new booking. When the mistake is noticed the new booking is ended and the prisoner is received again on the correct old booking. If the prisoner was re-measured on the new booking then DPS will have been updated but NOMIS loses this information. Therefore, whenever this happens we know DPS has the most up-to-date measurements so we synchronise from DPS back to NOMIS to keep NOMIS in line.


karenmillermoj (approved):
lgtm

PR #735, opened 2024-10-11
SDIT-2109 Pick up mapping api updates and start to add in child mapping

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/csip/CSIPService.kt
```diff
@@ -46,24 +49,39 @@ class CSIPService(
...
+              .let {
+                it.createNewMappings(dpsCsipReportId)
+              }
           }
       }
       saveMapping { mappingApiService.createMapping(it) }
     }
   }
 
+  fun UpsertCSIPResponse.createNewMappings(dpsCsipReportId: String) =
```
karenmillermoj, line 62: Not sure why I hung this off UpsertCSIPResponse - will change in another PR
All mappings are now passed through to the mapping service for the insert

petergphillips (approved):
LGTM

PR #763, opened 2024-10-28
SDIT-2143 Add extra fields to csip reconciliation

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/csip/CSIPReconciliationService.kt
```diff
@@ -51,20 +55,33 @@ class CSIPReconciliationService(
...
+    val missingFromDps = nomisCSIPs - dpsCSIPs
+    return if (mismatchCount != 0 || missingFromNomis.isNotEmpty() || missingFromDps.isNotEmpty()) {
+      MismatchCSIPs(
+        offenderNo = prisonerId.offenderNo,
+        dpsCSIPCount = dpsCSIPList.size,
+        nomisCSIPCount = nomisCSIPList.size,
+        missingFromDps = missingFromDps,
+        missingFromNomis = missingFromNomis,
+      ).also { mismatch ->
         log.info("CSIP Mismatch found  $mismatch")
```
petergphillips, line 75: Do you need to `log` and track the event?  Why do both - we're looking to keep our Log Analytics costs down so don't know if we need both?

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/csip/CSIPReconciliationService.kt
karenmillermoj: Yeah totally get that - in theory we should only get a few mismatches - if any, but I can totally see your point.

PR #768, opened 2024-10-30
SDIT-2172: 🐛 Ensure that futures are awaited for and avoid thenAccept

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/DomainEventListener.kt
```diff
@@ -76,9 +76,10 @@ abstract class DomainEventListener(
 
 private fun asCompletableFuture(
   process: suspend () -> Unit,
-): CompletableFuture<Void> = CoroutineScope(Context.current().asContextElement()).future {
+): CompletableFuture<Void?> = CoroutineScope(Context.current().asContextElement()).future {
   process()
-}.thenAccept { }
```
petergphillips, line 81: remove the `thenAccept`

mikehalmamoj (approved):

PR #771, opened 2024-11-01
SDIT-2181: ✨ Attempt to create PR if api changes

petergphillips:


build.gradle.kts
```diff
@@ -58,7 +58,7 @@ tasks {
       "buildMappingServiceApiModel",
       "buildSentencingAdjustmentsApiModel",
       "buildCourtSentencingApiModel",
-      "buildAlertApiModel",
+      "buildAlertsApiModel",
```
petergphillips, line 61: make it easier to translate between project and model build otherwise need a map of these as well

.github/workflows/check_api_docs.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Upload differences report
+        if: check-api.outputs.differences == 'true'
+        uses: actions/upload-artifact@v4
+        with:
+          path: build/api.diff
+
+      - name: Create Pull Request
+        if: check-api.outputs.differences == 'true'
+        uses: peter-evans/create-pull-request@v7
```
mikehalmamoj, line 36: Bit surpsied that a 3rd party action is required here - I would have thought Github would have their own action that just wraps the gh CLI or something.

mikehalmamoj (approved):

PR #813, opened 2024-12-03
SDIT-2317 Ignore attendance delete failed if not found

mikehalmamoj:
There are various reasons why an attendance might not exist in NOMIS but delete be requested by DPS. Having investigated these for at least year I've come to the conclusion that it doesn't matter - nothing bad has ever come of a failed attendance delete and the we just end up deleting the failed event from the DLQ.


petergphillips (approved):
LGTM

PR #814, opened 2024-12-03
SDIT-2336 endpoint to repair NOMIS alerts from DPS

andymarke:
* Gets all alerts in DPS (including inactive)
* Calls NOMIS API that will replace the alerts on the current booking
* Replaces the mappings for that prisoner

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsDataRepairResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsDataRepairResource.kt
@@ -1,0 +1,36 @@
+package uk.gov.justice.digital.hmpps.prisonertonomisupdate.alerts
+
+import com.microsoft.applicationinsights.TelemetryClient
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsDataRepairResource.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+class AlertsDataRepairResource(
+  private val alertsService: AlertsService,
+  private val telemetryClient: TelemetryClient,
+) {
+  @PostMapping("/prisoners/{offenderNo}/alerts/resynchronise")
+  @ResponseStatus(HttpStatus.OK)
+  @PreAuthorize("hasRole('ROLE_MIGRATE_ALERTS')")
```
mikehalmamoj, line 19: Is this the right role for this app?

mikehalmamoj (approved):

PR #828, opened 2024-12-09
SDIT-2309 contacts-api.contact-restriction.updated

andymarke:
and contacts-api.prisoner-contact-restriction.updated


petergphillips (approved):
LGTM

PR #838, opened 2024-12-20
SDIT-2301 contacts-api.contact-phone.deleted

andymarke:
and contacts-api.contact-address-phone.deleted


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/contactperson/ContactPersonDomainEventListener.kt
```diff
@@ -61,18 +63,21 @@ interface SourcedContactPersonEvent {
 }
 data class ContactCreatedEvent(
   override val additionalInformation: ContactAdditionalData,
-  val personReference: ContactIdentifiers,
-) : SourcedContactPersonEvent
+  override val personReference: ContactIdentifiers,
```
andymarke, line 66: ad interface so I can add the `contactId()` function

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/contactperson/ContactPersonService.kt
```diff
@@ -427,6 +427,32 @@ class ContactPersonService(
...
+    )
+
+    if (event.didOriginateInDPS()) {
+      mappingApiService.getByDpsContactPhoneIdOrNull(dpsContactPhoneId)?.nomisId?.also {
+        telemetryMap["nomisPhoneId"] = it.toString()
+        nomisApiService.deletePersonPhone(
+          personId = dpsContactId,
+          phoneId = it,
+        )
+        // TODO - Delete mapping - though behaviour wise all will work since the "from nomis" service will delete anyway
```
andymarke, line 447: I will add this in later, but it is not 100% necessary since the migration service process *all* deletes regardless of original (so it can handle cascading deletes) - so the mapping will be cleaned up anyway

mikehalmamoj (approved):

PR #841, opened 2024-12-31
SDIT-2285 Added mapping deletes

andymarke:
For each of the DPS deletes also delete the mapping

Also latest DPS OpenApi DTOs


petergphillips (approved):
LGTM

PR #846, opened 2025-01-07
SDIT-2384 Adjustments reconciliation use different paging strategy

andymarke:
Previous this uses the NOMIS api /prisoners/ids using Spring paging. This has now switched the "last id" /bookings/ids method

This should improve performance once we switch the all latest bookings rather than the current active bookings

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingReconciliationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingReconciliationService.kt
@@ -10,10 +10,10 @@ import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.config.trackEvent
+import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomissync.model.BookingIdsWithLast
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomissync.model.PrisonerIds
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.sentencing.adjustments.model.AdjustmentDto.AdjustmentType
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.NomisApiService
-import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.asPages
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.awaitBoth
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.doApiCallWithRetries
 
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingReconciliationService.kt
```diff
@@ -31,35 +31,52 @@ class SentencingReconciliationService(
...
 
-  suspend fun generateReconciliationReport(allPrisoners: Boolean, activePrisonersCount: Long): List<MismatchSentencingAdjustments> = activePrisonersCount.asPages(pageSize).flatMap { page ->
-    val activePrisoners = getActivePrisonersForPage(allPrisoners, page)
+  suspend fun generateReconciliationReport(allPrisoners: Boolean): List<MismatchSentencingAdjustments> {
+    val mismatches: MutableList<MismatchSentencingAdjustments> = mutableListOf()
+    var lastBookingId = 0L
 
-    withContext(Dispatchers.Unconfined) {
-      activePrisoners.map { async { checkBookingAdjustmentsMatch(it) } }
+    do {
```
andymarke, line 38: replaced bu do..while - keeping calling page end point until we get to the last page

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingReconciliationService.kt
```diff
@@ -31,35 +31,52 @@ class SentencingReconciliationService(
...
-            "page" to page.first.toString(),
+            "booking" to lastBookingId.toString(),
           ),
         )
-        log.error("Unable to match entire page of prisoners: $page", it)
+        log.error("Unable to match entire page of bookings from booking: $lastBookingId", it)
       }
-      .getOrElse { emptyList() }
-      .also { log.info("Page requested: $page, with ${it.size} active prisoners") }
+      .getOrNull()
```
andymarke, line 78: we need a way of indicating that this call failed so just continue with next page - using null to indicate that.

I might re-think this - it is a bit shoddy

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingResource.kt
```diff
@@ -30,16 +28,10 @@ class SentencingResource(
   @PutMapping("/sentencing/reports/reconciliation")
   @ResponseStatus(HttpStatus.ACCEPTED)
   suspend fun generateSentencingReconciliationReport() {
-    val prisonerCount = nomisApiService.getActivePrisoners(
-      pageNumber = 0,
-      pageSize = 1,
-    ).totalElements
-
-    telemetryClient.trackEvent("sentencing-reports-reconciliation-requested", mapOf("prisoner-count" to prisonerCount.toString()))
```
andymarke, line 38: can no longer output the expected count

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/sentencing/SentencingResourceIntTest.kt
```diff
@@ -153,18 +177,12 @@ class SentencingResourceIntTest : IntegrationTestBase() {
       )
     }
 
-    @Test
-    fun `when initial prison count fails the whole report fails`() {
```
andymarke, line 157: no longer true, it will just ignore a first page failure

PR #853, opened 2025-01-14
SDI-1855-Casenotes-reconciliation-tweaks---mapping-retries

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -226,7 +226,7 @@ data class CommonCaseNoteFields(
 ) {
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
-    other as CommonCaseNoteFields
+    if (other !is CommonCaseNoteFields) return false
```
steverendell, line 229: fixes other being null

lbennett-moj (approved):

PR #856, opened 2025-01-15
SDIT-2419 Increase heap size

mikehalmamoj:


helm_deploy/hmpps-prisoner-to-nomis-update/values.yaml
```diff
@@ -4,6 +4,14 @@ generic-service:
 
   replicaCount: 4
 
+  resources:
+    requests:
+      cpu: 10m
+      memory: 640Mi
+    limits:
+      cpu: 2000m
+      memory: 2048Mi
```
mikehalmamoj, line 13: This overrides the limit range defined in the Cloud Platform terraform, which is currently set to a max of 1024Mi. As by default the heap gets 1/4 of the max, this explains why the the heap is always >90% full for this app.

We could explicitly set the size of all JVM memory components including the heap as explained in [this article](https://cloud.theodo.com/en/blog/jvm-oom). But 1/4 seems fairly sensible to me and it's the default for a reason.

helm_deploy/hmpps-prisoner-to-nomis-update/values.yaml
```diff
@@ -75,7 +83,6 @@ generic-service:
 
   # Environment variables to load into the deployment
   env:
-    JAVA_OPTS: "-Xmx512m"
```
mikehalmamoj, line 78: This does nothing so remove to avoid confusion.

petergphillips (approved):
LGTM

PR #859, opened 2025-01-21
SDI-2479 Case notes reconciliation amendment times different precision

steverendell:
DPS timestamps contain nanos whereas Nomis ones just seconds. Reconciliation needs to account for this


petergphillips (approved):
LGTM

PR #861, opened 2025-01-22
SDI-2481 Case notes reconciliation username check can fail

steverendell:
Take all usernames for the staff id into account


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -241,17 +244,17 @@ data class CommonCaseNoteFields(
   private fun equalTypes(nomis: CommonCaseNoteFields): Boolean = (type == nomis.type || (type == "APP" && nomis.type == "CNOTE" && subType == "OUTCOME"))
 
   private fun equalUsers(other: CommonCaseNoteFields): Boolean {
-    val equal = authorUsername == other.authorUsername || authorUsername == "OMS_OWNER"
+    val equal = other.nomisUsernames?.contains(dpsUsername) == true || dpsUsername == "OMS_OWNER"
```
steverendell, line 247: This is not ideal as it is asymetrical - though I am thinking of splitting this out to report what specific field does not match, when there is time

PR #868, opened 2025-01-31
SDIT-2541 Increase max response size

mikehalmamoj:


src/main/resources/application.yml
```diff
@@ -6,7 +6,7 @@ spring:
   application:
     name: hmpps-prisoner-to-nomis-update
   codec:
-    max-in-memory-size: 10MB
+    max-in-memory-size: 15MB
```
mikehalmamoj, line 9: This is a workaround - we need to get the response size down from the API we call, or choose a different endpoint or something like that.

andymarke (approved):
lgtm

PR #869, opened 2025-01-31
SDIT-2544: ✨ Allow reconciliation to be run on an individual prisoner

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -103,72 +103,75 @@ class CaseNotesReconciliationService(
...
+      null
+    } else {
       telemetryClient.trackEvent(
-        "casenotes-reports-reconciliation-mismatch-error",
+        "casenotes-reports-reconciliation-mismatch",
         mapOf(
           "offenderNo" to offenderNo,
-          "error" to (it.message ?: ""),
-        ),
+        ) + mismatchCaseNote.notes.withIndex().associateBy({ (it.index + 1).toString() }, { it.value }),
```
petergphillips, line 171: was a bit naff to have a i variable here, so removed it

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivityResourceIntTest.kt
```diff
@@ -203,7 +203,7 @@ class ActivityResourceIntTest : IntegrationTestBase() {
 
     @Test
     fun `will return error if anything fails`() = runTest {
-      activitiesApi.stubGetScheduleWithError(ACTIVITY_SCHEDULE_ID)
+      mappingServer.stubGetMappingsWithError(ACTIVITY_SCHEDULE_ID, 503)
```
petergphillips, line 206: it was expecting the wrong API call first so actually triggered a `404` instead, which we now map to not found

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesResource.kt
```diff
@@ -59,4 +68,57 @@ class CaseNotesResource(
...
+        ],
+      ),
+    ],
+  )
+  suspend fun generateReconciliationReportForPrisoner(
+    @Schema(description = "Prison number aka noms id / offender id display", example = "A1234BC") @PathVariable prisonNumber: String,
+  ) = try {
+    caseNotesReconciliationService.checkMatchOrThrowException(prisonNumber)
+  } catch (notFound: NotFound) {
+    throw NotFoundException("Offender not found $prisonNumber")
```
petergphillips, line 120: now throwing own exception that I've added to the handler

mikehalmamoj (approved):

PR #874, opened 2025-02-03
SDIT-2547: ✨ Improve mismatch case note output

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -177,40 +177,40 @@ class CaseNotesReconciliationService(
...
-    text.truncateToFixNomisMaxLength(),
-    type,
-    subType,
-    occurrenceDateTime.format(truncatedToSecondsFormatter),
-    creationDateTime.format(truncatedToSecondsFormatter),
-    authorUsername,
-    null,
-    amendments.map { a ->
+  private fun CaseNote.transformFromDps() = ComparisonCaseNote(
+    id = caseNoteId,
```
petergphillips, line 181: add in the case note id so that we can easily go to DPS to get the case note.  At present we have to get all the case notes for a prisoner and then look for the one we need

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -226,23 +226,24 @@ data class MismatchCaseNote(
...
+data class ComparisonCaseNote(
+  val id: String,
   val text: String,
   val type: String,
   val subType: String,
   val occurrenceDateTime: String?,
   val creationDateTime: String?,
-  val dpsUsername: String?,
-  val nomisUsernames: List<String>?,
+  val dpsUsername: String? = null,
```
petergphillips, line 239: default to `null` for usernames so that we don't need to pass in `null` where it doesn't make sense

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -226,23 +226,24 @@ data class MismatchCaseNote(
   val offenderNo: String,
   val diffsForDps: Set<String> = emptySet(),
   val diffsForNomis: Set<Long> = emptySet(),
-  var notes: List<String> = emptyList(),
+  val notes: List<String> = emptyList(),
 )
 
-data class CommonCaseNoteFields(
+data class ComparisonCaseNote(
```
petergphillips, line 232: renamed as these weren't really common fields before I added the ids 😁.  It is more a class that is used to compare two case notes

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -253,19 +254,19 @@ data class CommonCaseNoteFields(
       equalAmendments(this, other)
   }
 
-  private fun equalTypes(nomis: CommonCaseNoteFields): Boolean = (type == nomis.type || (type == "APP" && nomis.type == "CNOTE" && subType == "OUTCOME"))
+  private fun equalTypes(nomis: ComparisonCaseNote): Boolean = type == nomis.type || (type == "APP" && nomis.type == "CNOTE" && subType == "OUTCOME")
 
-  private fun equalUsers(other: CommonCaseNoteFields): Boolean {
-    val equal = other.nomisUsernames?.contains(dpsUsername) == true || dpsUsername == "OMS_OWNER"
+  private fun equalUsers(nomis: ComparisonCaseNote): Boolean {
```
petergphillips, line 259: slightly confusing here, but the `equals` isn't symmetric as it always assumes DPS is on the left hand side of the comparison.  This duplicates the `equalTypes` method approach where the parameter name of `nomis` reflects that the nomis case note is on the right hand side

mikehalmamoj (approved):

PR #875, opened 2025-02-04
SDIT-2547: 🐛 Combine amendments into text for comparison

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -177,48 +177,53 @@ class CaseNotesReconciliationService(
...
-    id = caseNoteId,
-    text = text.truncateToFixNomisMaxLength(),
-    type = type,
-    subType = subType,
-    occurrenceDateTime = occurrenceDateTime.format(truncatedToSecondsFormatter),
-    creationDateTime = creationDateTime.format(truncatedToSecondsFormatter),
-    dpsUsername = authorUsername,
-    amendments = amendments.map { a ->
+  private fun CaseNote.transformFromDps(): ComparisonCaseNote {
+    val sortedAmendments = amendments.map { a ->
```
petergphillips, line 181: not sure why we need to sort both lots of amendments, would have thought that they would arrive in the right order...

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -238,7 +243,6 @@ data class ComparisonCaseNote(
   val creationDateTime: String?,
   val dpsUsername: String? = null,
   val nomisUsernames: List<String>? = null,
-  val amendments: Set<CommonAmendmentFields>,
```
petergphillips, line 241: amendments not now used in comparison since DPS can have more amendments than NOMIS due to the NOMIS text holding all the amendments.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -275,6 +277,7 @@ data class CommonAmendmentFields(
   val authorUsername: String,
 ) {
   override fun toString(): String = "{text-hash=${Objects.hashCode(text)}, occurrenceDateTime=$occurrenceDateTime, authorUsername=$authorUsername}"
+  fun toNomisAmendment(): String = " ...[$authorUsername updated the case notes on $occurrenceDateTime] $text"
```
petergphillips, line 280: we copy the amendment into the case note text so are therefore testing all the amendment fields here anyway

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -177,48 +177,53 @@ class CaseNotesReconciliationService(
...
         occurrenceDateTime = a.creationDateTime?.format(truncatedToSecondsFormatter),
         authorUsername = a.authorUserName,
       )
-    }.toSortedSet(amendmentComparator),
-    legacyId = legacyId,
-  )
+    }.toSortedSet(amendmentComparator)
+    return ComparisonCaseNote(
+      id = caseNoteId,
+      text = (text + sortedAmendments.joinToString("") { it.toNomisAmendment() }).truncateToFixNomisMaxLength(),
```
mikehalmamoj, line 190: Does this truncate mean that you don't include amendments in the comparison if the text is already max size?

karenmillermoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
petergphillips: yes, exactly. That is what is happening at present though.  See https://prisoner-dev.digital.prison.service.justice.gov.uk/prisoner/G3101UO/case-notes for an example.  The amendments are missed off in the NOMIS database and only shown in DPS.

PR #892, opened 2025-02-12
SDIT-2563: 🐛 Have a go at getting counts for dps and nomis to match

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/nonassociations/NonAssociationsReconciliationService.kt
```diff
@@ -29,25 +29,24 @@ class NonAssociationsReconciliationService(
   private val pageSize: Long,
 ) {
 
-  var nomisTotalDetails = 0
```
petergphillips, line 32: private variable in the service that was keeping a count of totals across `async` threads

andymarke (approved):
lgtm

PR #906, opened 2025-02-25
SDIT-2617 Adjudication repair endpoint.

andymarke:
Set punishment date to latest hearing date when not present.

Since this repairs a historic adjudication the punishment date would normally be "now" which is when the hearing is entered, but in this case it makes more sense to be the actual date the hearing took place, else a hearing for 10 years ago would have a punishment date of today

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
@@ -158,13 +158,10 @@ class AdjudicationsService(
       )
     }
     createPunishments(
-      PunishmentEvent(
-        PunishmentsAdditionalInformation(
-          chargeNumber = chargeNumber,
-          prisonId = prisonId,
-          prisonerNumber = offenderNo,
-        ),
-      ),
+      chargeNumber = chargeNumber,
+      prisonId = prisonId,
+      prisonerNumber = offenderNo,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -1015,10 +1025,10 @@ class AdjudicationsService(
 
   private inline fun <reified T> String.fromJson(): T = objectMapper.readValue(this)
 
-  private suspend fun PunishmentDto.toNomisAward() = HearingResultAwardRequest(
+  private suspend fun PunishmentDto.toNomisAward(hearingDate: LocalDate? = null) = HearingResultAwardRequest(
     sanctionType = this.type.toNomisSanctionType(),
     sanctionStatus = this.toNomisSanctionStatus(),
-    effectiveDate = this.schedule.startDate ?: this.schedule.suspendedUntil ?: LocalDate.now(),
+    effectiveDate = this.schedule.startDate ?: this.schedule.suspendedUntil ?: hearingDate ?: LocalDate.now(),
     sanctionDays = this.schedule.takeIf { it.measurement == PunishmentScheduleDto.Measurement.DAYS }?.duration ?: 0,
```
andymarke, line 1032: only use the hearing date if DPS does not provide a schedule (e.g not suspended or has a real start date)

petergphillips (approved):
LGTM

PR #917, opened 2025-02-27
SDIT-2635: 🐛 Case note amendment truncation fix

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -201,7 +203,7 @@ class CaseNotesReconciliationService(
     val sortedAmendments = amendments.map { a ->
       CommonAmendmentFields(
         text = a.text,
-        occurrenceDateTime = a.createdDateTime,
+        occurrenceDateTime = LocalDateTime.parse(a.createdDateTime).format(amendmentFormatter),
```
petergphillips, line 206: this is really naff. The API generation generates this field as a string rather than a date time, so we need to parse it first

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -176,12 +177,13 @@ class CaseNotesReconciliationService(
...
 
   private val truncatedToSecondsFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss")
+  private val amendmentFormatter = DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm:ss")
 
   private fun CaseNote.transformFromDps(): ComparisonCaseNote {
     val sortedAmendments = amendments.map { a ->
       CommonAmendmentFields(
         text = a.additionalNoteText,
-        occurrenceDateTime = a.creationDateTime?.format(truncatedToSecondsFormatter),
+        occurrenceDateTime = a.creationDateTime?.format(amendmentFormatter),
```
petergphillips, line 186: by using the modern (post 2018) amendment format it means that if we encounter an amendment that is truncated (due to case note & amendments each having a 4,000 limit in DPS) then we can still compare

mikehalmamoj (approved):

PR #921, opened 2025-02-28
SDIT-2628 Building a more functionally organisation reconciliation se…

andymarke:
…rvice

TODO
- address phones

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/organisations/OrganisationsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/organisations/OrganisationsResource.kt
@@ -15,7 +15,6 @@ import uk.gov.justice.digital.hmpps.prisonertonomisupdate.config.trackEvent
 class OrganisationsResource(
   private val telemetryClient: TelemetryClient,
   private val reconciliationService: OrganisationsReconciliationService,
-  private val nomisApiService: OrganisationsNomisApiService,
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/organisations/OrganisationsResource.kt
```diff
@@ -26,15 +25,25 @@ class OrganisationsResource(
...
   @ResponseStatus(HttpStatus.ACCEPTED)
   suspend fun generateOrganisationsReconciliationReport() {
-    val organisationsCount = nomisApiService.getCorporateOrganisationIds(0, 1).totalElements
+    val counts = reconciliationService.getOrganisationsCounts()
 
-    telemetryClient.trackEvent("organisations-reports-reconciliation-requested", mapOf("organisations" to organisationsCount.toString()))
+    telemetryClient.trackEvent("organisations-reports-reconciliation-requested", mapOf("organisations" to counts.nomisCount.toString()))
+
+    if (counts.nomisCount != counts.dpsCount) {
+      telemetryClient.trackEvent(
```
andymarke, line 33: Checks that a DPS doesn't have an orphaned organisation - this is probably good enough, but we might need to improve if this ever happens and we need to fix the one(s) that are missing.
But since there is no sync from DPS right now we can worry about that if it ever happens

lbennett-moj (approved):

PR #923, opened 2025-03-03
SDIT-2639: ♻️ Use new nomis prisoner api that uses LocalDateTime

petergphillips:
- **SDIT-2639: ♻️ Use new nomis prisoner api that uses LocalDateTime**


karenmillermoj (approved):
lgtm

PR #926, opened 2025-03-04
SDIT-2628 Building a more functionally organisation reconciliation se…

andymarke:
…rvice

Address phones. Addresses are hard to compare given nothing is mandatory for an address and an organisations often have lots of identical empty addresses; so this just compares the set of phone numbers


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/organisations/OrganisationsResource.kt
```diff
@@ -33,8 +33,8 @@ class OrganisationsResource(
       telemetryClient.trackEvent(
         "organisations-reports-reconciliation-mismatch",
         mapOf(
-          "dpsCount" to "$counts.dpsCount",
-          "nomisCount" to "$counts.nomisCount",
+          "dpsCount" to "${counts.dpsCount}",
```
andymarke, line 36: silly mistake that the rubbish test  didn't find. Corrected both

mikehalmamoj (approved):

PR #942, opened 2025-03-06
pgp SDIT 2181 check docs action

petergphillips:
- **SDIT-2181: ✨ Initial stab at github action to compare differences**
- **SDIT-2181: ✨ Initial stab at github action to compare differences**


.github/workflows/check_api_docs.yml
```diff
@@ -0,0 +1,53 @@
...
+          - personal-relationships
+          - sentencing-adjustments
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  build:
+    runs-on: [ self-hosted, hmpps-github-actions-runner ]
```
petergphillips, line 31: this line means it will run in cloud platform, so will have access to the dev servers

andymarke (approved):
lgtm

PR #953, opened 2025-03-10
SDIT-2586 Handle switch to old booking edge case

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/profiledetails/ContactPersonProfileDetailsSyncService.kt
```diff
@@ -106,6 +109,21 @@ class ContactPersonProfileDetailsSyncService(
...
+    val prisonerNumber = event.additionalInformation.nomsNumber
+
+    if (eventFeatureSwitch.isEnabled("personal-relationships-api.domestic-status.created")) {
+      syncProfileDetail(prisonerNumber, 0, MARITAL)
+    }
+
+    if (eventFeatureSwitch.isEnabled("personal-relationships-api.number-of-children.created")) {
+      syncProfileDetail(prisonerNumber, 0, CHILD)
+    }
+  }
```
mikehalmamoj, line 125: The main change is to listen for prisoner received events with reason of switch booking, then sync everything back to NOMIS.

PR #954, opened 2025-03-10
SDIT-2181: 👽️ Grab latest non associations and try to restrict models

petergphillips:


build.gradle.kts
```diff
@@ -85,11 +85,13 @@ val models = listOf(
     name = "non-associations",
     packageName = "nonassociations",
     url = "https://non-associations-api-dev.hmpps.service.justice.gov.uk/v3/api-docs",
+    models = "LegacyNonAssociation,LegacyNonAssociationOtherPrisonerDetails,NonAssociation",
```
petergphillips, line 88: for old projects that aren't in active development try to constrain list of models so that we only pick up differences if these change

ushkarev: @petergphillips why do you want to use the legacy models? does the newer one not contain quite what you're after?

PR #956, opened 2025-03-11
SDIT-2181: 👽️ Grab latest adjudications and try harder to restrict models

petergphillips:


build.gradle.kts
```diff
@@ -183,6 +187,18 @@ models.forEach {
       Files.write(Paths.get(it.input), formattedJson.toByteArray())
     }
   }
+  tasks.register(it.toReadProductionVersionTaskName()) {
```
petergphillips, line 190: not sure this is going to be useful, but have added it for now

mikehalmamoj (approved):

PR #977, opened 2025-03-14
SDIT-2494 Add sequence to reviewrequest and add repair endpoint

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/csip/CSIPService.kt
```diff
@@ -187,6 +187,17 @@ class CSIPService(
     log.warn("Unable to delete mapping for csip $dpsCsipId. Please delete manually", e)
   }
 
+  suspend fun repairCsipReport(prisonNumber: String, dpsCsipReportId: String) = updateCSIPReport(
+    CSIPEvent(
```
karenmillermoj, line 191: The repair endpoint is for a basic repair - so updates Nomis with the latest data

mikehalmamoj (approved):

PR #986, opened 2025-03-18
SDIT-2628 Contact reconciliation

andymarke:
* Check restriction dates
* Cope with duplicate contacts


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonReconciliationService.kt
```diff
@@ -93,18 +94,8 @@ class ContactPersonReconciliationService(
         async { dpsApiService.getPrisonerContacts(prisonerId.offenderNo).content!! }
     }.awaitBoth()
 
-    val dpsContactSummaries = dpsContacts.map { it.asSummary() }.sortedWith(
```
andymarke, line 96: abandoned the idea of ordering and compare pairs - it doesn't work when there are duplicates like there are In live due to bad data

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonReconciliationService.kt
```diff
@@ -279,12 +258,16 @@ class ContactPersonReconciliationService(
     contactId = contactId,
     relationshipCode = relationshipCode,
     restrictionType = this.type.code,
+    startDate = this.effectiveDate,
+    expiryDate = this.expiryDate,
   )
```
andymarke, line 263: now checking restriction dates that implicitly checks the status

mikehalmamoj (approved):

PR #997, opened 2025-03-21
SDIT-2570 Adjust reconciliation job times

mikehalmamoj:


helm_deploy/values-dev.yaml
```diff
@@ -92,3 +92,6 @@ generic-prometheus-alerts:
     organisations:
       - "syscon-devs-dev-hmpps_prisoner_to_nomis_organisations_queue"
       - "syscon-devs-dev-hmpps_prisoner_to_nomis_organisations_dlq"
+
+cron:
+  contactPersonProfileDetailsReportSchedule: "30 7 * * *"
```
mikehalmamoj, line 97: There's not much point in running this at 3.30 when nothing is up so it just errors.

andymarke: Interestingly I never run reconciliation in dev other to test once

andymarke (approved):
lgtm

helm_deploy/values-dev.yaml
mikehalmamoj: I'll see how it goes. Initially dev was a useful canary for problems with activities sync - but after a while it just became noise.

PR #999, opened 2025-03-21
SDIT-2679 Ignore deleted adjustments when already deleted

andymarke:
* Scenario is NOMIS apis hangs but delete succeeds
* migration picks up the delete event from NOMIS and deletes mapping
* endless retry loop since mapping is no long found even though record has been deleted


petergphillips (approved):
LGTM

PR #1015, opened 2025-03-26
SDIT-2613 allow null comparison for visit balance reconciliation and add prisoner reconciliation

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/visitbalances/VisitBalanceResource.kt
```diff
@@ -29,19 +37,74 @@ class VisitBalanceResource(
...
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+    ],
+  )
+  suspend fun generateReconciliationReportForPrisoner(
+    @Schema(description = "Prison number aka noms id / offender id display", example = "A1234BC")
+    @PathVariable prisonNumber: String,
```
karenmillermoj, line 108: May want a different result (than match) if offender or visit balance doesn't exist in both Nomis and Dps - will do in separate PR

PR #1047, opened 2025-04-07
SDIT-2704 Handle duplicate case notes from to-nomis sync - fix

steverendell:
Find DPS case note by contents, not by legacy id

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
@@ -182,8 +182,13 @@ class CaseNotesReconciliationService(
   private val maxLengthToCompare = MAX_CASENOTE_LENGTH_BYTES - SEE_DPS_REPLACEMENT.length
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/casenotes/CaseNotesReconciliationService.kt
```diff
@@ -197,24 +202,34 @@ class CaseNotesReconciliationService(
...
+          val nomisCaseNote2 = identicalCaseNotePair.last()
+          val textToCompare = nomisCaseNote1.text.take(maxLengthToCompare)
+          val exactlyOneExistsInDps = dpsCaseNotes.values.filter {
+            it.text.take(maxLengthToCompare) == textToCompare &&
+              it.occurrenceDateTime == nomisCaseNote1.occurrenceDateTime &&
+              it.creationDateTime == nomisCaseNote1.creationDateTime &&
+              it.type == nomisCaseNote1.type &&
+              it.subType == nomisCaseNote1.subType
+          }
+            .size == 1
```
steverendell, line 215: Altered logic to find DPS case note using details as the legacy id is deprecated and cannot be relied on

andymarke (approved):
lgtm

PR #1053, opened 2025-04-08
SDIT-2708 Ignore deleted hearings for a deleted hearing outcome

andymarke:
Fixes issue for false positives on DLQ when a hearing and outcome are deleted in quick succession


steverendell (approved):
lgtm

PR #1060, opened 2025-04-09
SDIT-2373 Config person and prisoner reconciliation

andymarke:
Person config has a lower demand on DPS (though will take much longer) so can have more threads


karenmillermoj (approved):
lgtm

PR #1061, opened 2025-04-09
SDIT-2373 Move Person reconciliation to use channels

andymarke:
This now uses the coroutines more efficiently.
Previously the threads would call to DPS/NOMIS in waves and wait for each batch to complete. This will have a constant "pool" of routines always feeding of the available personIds. Overall time should be reduced

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonReconciliationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonReconciliationService.kt
@@ -1,9 +1,18 @@
 package uk.gov.justice.digital.hmpps.prisonertonomisupdate.personalrelationships
 
 import com.microsoft.applicationinsights.TelemetryClient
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonReconciliationService.kt
```diff
@@ -74,7 +85,26 @@ class ContactPersonReconciliationService(
...
+        mismatchesChannel.close()
+      }
+      mismatchesChannel.toList()
+    }
+  }
+
+  fun CoroutineScope.launchCheckPersonMatcher(channel: ReceiveChannel<Long>, mismatchesChannel: Channel<MismatchPersonContacts>) = launch {
+    for (personId in channel) {
+      checkPersonContactMatch(personId)?.also { mismatchesChannel.send(it) }
+    }
```
andymarke, line 104: if there is a mismatch send to mismatch channel rather than returning

petergphillips (approved):
LGTM

PR #1070, opened 2025-04-14
SDIT-2373 Reduce to 30 reconciliation person threads

andymarke:
Should take 4 hours with less stress on DB


petergphillips (approved):
LGTM

PR #1117, opened 2025-05-02
SDIT-2710 Pick up visit balance adjustment response dto from visits api

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/data/DomainEvents.kt
```diff
@@ -0,0 +1,33 @@
+package uk.gov.justice.digital.hmpps.prisonertonomisupdate.data
+
+import java.time.OffsetDateTime
+
+data class PrisonerMergeDomainEvent(
```
karenmillermoj, line 5: Copied over from "from-nomis-migration" - @petergphillips if we don't need these, we can just remove them

petergphillips (approved):
LGTM

PR #1136, opened 2025-05-09
SDIT-2762 Allow data repair to clear child mappings

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/csip/CSIPDataRepairResource.kt
```diff
@@ -25,8 +27,10 @@ class CSIPDataRepairResource(
   suspend fun repairCsipReport(
     @PathVariable prisonNumber: String,
     @PathVariable dpsCsipReportId: String,
+    @Schema(description = "Whether to clear existing child mappings", required = false, defaultValue = "false")
+    @RequestParam clearChildMappings: Boolean = false,
```
karenmillermoj, line 31: It is very rare to have to tidy up mappings.  It seems to occur when 2 messages for the same report get entangled. With further changes in nomis-prisoner-api to set up locking on the db, this should never be needed.

PR #1145, opened 2025-05-12
SDIT-2754 Handle errors found in preprod caused by preprod refresh

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesReconService.kt
```diff
@@ -161,7 +162,7 @@ class ActivitiesReconService(
     differences.all()
       .filterNot { it in ignoredBookings }
       .sorted()
-      .map { differencesWithDetails.first { details -> details.bookingId == it } }
+      .map { differencesWithDetails.firstOrNull { details -> details.bookingId == it } ?: PrisonerDetails("Details not found in NOMIS", it, prisonId, false) }
```
mikehalmamoj, line 165: Preprod has some bookings in activities but not in NOMIS due to differences in time of preprod refresh. Seems like a reasonable idea to fix this just in case the same problem ever happens in production, e.g. a booking is deleted from NOMIS but not DPS activities.

karenmillermoj (approved):
lgtm

PR #1146, opened 2025-05-12
SDIT-2413: ✨ Add in calls to DPS sync endpoints

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/organisations/OrganisationsDpsApiService.kt
```diff
@@ -36,4 +43,39 @@ class OrganisationsDpsApiService(
     }
     .retrieve()
     .awaitBodyWithRetry(backoffSpec)
+
+  suspend fun getSyncOrganisation(organisationId: Long): SyncOrganisationResponse = webClient.get()
+    .uri("/sync/organisation/{organisationId}", organisationId)
+    .retrieve()
+    .awaitBodyWithRetry(backoffSpec)
```
petergphillips, line 50: I presume we want this rather than just an `.awaitBody()`.  Am copying from [ContactPersonDpsApiService](https://github.com/ministryofjustice/hmpps-prisoner-to-nomis-update/blob/262566a3f4cc42645d2d81738392ec86e43f82fc/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonDpsApiService.kt#L35) and that just has `awaitBody` instead

mikehalmamoj (approved):

PR #1170, opened 2025-05-20
SDIT-2776 Purge the activities DLQ in preprod before the retryDLQ job runs

mikehalmamoj:
In preprod the DPS activities database and NOMIS sync mappings databases are refreshed from prod several hours after the NOMIS preprod database is refreshed. This means data is out of sync and causes lots of events published by activities to fail. These failures are then retried every 10 minutes but always fail. As we rollout to more prisons this has turned into 10s of thousands of failing DLQ messages every 10 minutes. This is very noisy and achieves nothing.

A proper fix will involve refreshing all DPS databases at the same time and making sure all DPS systems are down until the refresh has completed. This will be complicated.

As a medium term workaround we will instead purge the DLQ before each retry takes place to stop the noise.

### DIFF
--- a/helm_deploy/hmpps-prisoner-to-nomis-update/templates/activities-purge-dlq-cronjob.yaml
+++ b/helm_deploy/hmpps-prisoner-to-nomis-update/templates/activities-purge-dlq-cronjob.yaml
@@ -1,0 +1,35 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "app.fullname" . }}-activities-purge-dlq
+  labels:
+    {{- include "app.labels" . | nindent 4 }}
+spec:
+  schedule: "9-59/10 * * * *"
+  concurrencyPolicy: Forbid
+  failedJobsHistoryLimit: 5
+  startingDeadlineSeconds: 600
+  successfulJobsHistoryLimit: 5
+  jobTemplate:
+    spec:
+      # Tidy up all jobs after 4 days
+      ttlSecondsAfterFinished: 345600
+      suspend: {{ .Values.cron.activitiesPurgeDlqSuspend }}
+      template:
+        spec:
+          containers:
+            - name: housekeeping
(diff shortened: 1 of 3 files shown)


helm_deploy/hmpps-prisoner-to-nomis-update/templates/activities-purge-dlq-cronjob.yaml
```diff
@@ -0,0 +1,35 @@
...
+      suspend: {{ .Values.cron.activitiesPurgeDlqSuspend }}
+      template:
+        spec:
+          containers:
+            - name: housekeeping
+              image: ghcr.io/ministryofjustice/hmpps-devops-tools
+              args:
+                - /bin/sh
+                - -c
+                - curl --retry 2 -XPUT http://hmpps-prisoner-to-nomis-update/queue-admin/purge-queue/syscon-devs-preprod-hmpps_prisoner_to_nomis_activity_dlq
```
mikehalmamoj, line 26: I've hardcoded the preprod DLQ name here because I don't envisage this being needed for any other environments at the moment, so no config needed per env.

helm_deploy/hmpps-prisoner-to-nomis-update/templates/activities-purge-dlq-cronjob.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "app.fullname" . }}-activities-purge-dlq
+  labels:
+    {{- include "app.labels" . | nindent 4 }}
+spec:
+  schedule: "9-59/10 * * * *"
```
mikehalmamoj, line 8: The retryDLQ job runs every 10 minutes from 00 so we clear the DLQ prior to this.

lbennett-moj (approved):

PR #1172, opened 2025-05-21
SDIT-2764 Convert to Recall

andymarke:
* Call improved NOMIS API endpoint that passes all sentence types
* Call legacy version of /recall from DPS


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -820,36 +821,26 @@ class CourtSentencingService(
...
+                sentenceCalcType = dpsSentence.sentenceCalcType,
               )
             },
-            returnToCustody = recall.returnToCustodyDate?.let {
+            returnToCustody = recall.returnToCustodyDate?.takeIf { recall.recallType.isFixedTermRecall() }?. let {
               ReturnToCustodyRequest(
                 returnToCustodyDate = it,
-                enteredByStaffUsername = recall.createdByUsername,
-                // TODO we should only do this if a FTR - so check if returnToCustodyDate from DPS determines this rather then mapping to 0 days
-                recallLength = when (recall.recallType) {
```
andymarke, line 843: extracted to function

PR #1180, opened 2025-05-22
SDIT-2784 Update recall sentences

andymarke:
* Does the exact same as a create which is possibly ok, needs some testing
* Small workaround for the domain event not containing the sentence Ids

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
@@ -860,6 +860,73 @@ class CourtSentencingService(
       telemetryClient.trackEvent("recall-inserted-ignored", telemetryMap, null)
     }
   }
+  suspend fun updateRecallSentences(recallUpdateEvent: RecallEvent) {
+    val source = recallUpdateEvent.additionalInformation.source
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -860,6 +860,73 @@ class CourtSentencingService(
       telemetryClient.trackEvent("recall-inserted-ignored", telemetryMap, null)
     }
   }
+  suspend fun updateRecallSentences(recallUpdateEvent: RecallEvent) {
```
andymarke, line 863: this is nearly identical to the inserted above exact how it retrieves sentence Ids - might collapses these into a single function fi testing shows that can really do the same thing

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -1116,6 +1183,7 @@ class CourtSentencingService(
 
   data class RecallAdditionalInformation(
     val recallId: String,
+    val previousRecallId: String?,
     val sentenceIds: List<String>,
```
andymarke, line 1187: not used yet, but will be for deletes

lbennett-moj (approved):

PR #1188, opened 2025-05-23
SDIT-2783 Process recall.deleted

andymarke:
* Also refactored to reduce duplicate code
* TODO - create specific NOMIS API calls


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -927,6 +909,113 @@ class CourtSentencingService(
...
+    val telemetryMap = mutableMapOf(
+      "dpsRecallId" to recallId,
+      "offenderNo" to offenderNo,
+      "dpsSentenceIds" to recallDeletedEvent.additionalInformation.sentenceIds.joinToString(),
+    )
+
+    if (isDpsCreated(source)) {
+      runCatching {
+        if (previousRecallId != null) {
+          telemetryMap["dpsPreviousRecallId"] = previousRecallId
```
andymarke, line 926: If there is a previous recall this is more or less just like another update

lbennett-moj (approved):

PR #1206, opened 2025-05-28
SDIT-2778 Switch to DPS Contacts reconcile endpoint

andymarke:
Reduces the number of calls to DPS by using a specific crafted reconcile endpoint

Also cleaned some IntelliJ warnings including grammar warnings


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonDpsApiService.kt
```diff
@@ -79,13 +78,8 @@ class ContactPersonDpsApiService(@Qualifier("personalRelationshipsApiWebClient")
     .retrieve()
     .awaitBody()
 
-  suspend fun getPrisonerContacts(prisonNumber: String): PagedModelPrisonerContactSummary = webClient.get()
```
andymarke, line 82: These 2 calls are replaced with the single call below

lbennett-moj (approved):

PR #1210, opened 2025-05-29
SDIT-2778 Switch to Channels for Prisoner contact reconciliation

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonReconciliationService.kt
```diff
@@ -54,33 +52,25 @@ class ContactPersonReconciliationService(
     const val TELEMETRY_PERSON_PREFIX = "contact-person-reconciliation"
   }
 
-  suspend fun generatePrisonerContactReconciliationReport(): List<MismatchPrisonerContacts> {
```
andymarke, line 57: Essentially this is the same as person reconciliation

PR #1218, opened 2025-05-30
SDIT-2778 Refactor to extract channel logic

andymarke:
SDIT-2778 Refactor to extract channel logic

Extract all Channel and paging logic to ReconciliationLongPagingHelper.kt file.

The helper is typed, though only supports the paging pattern that returns a last ID that will be supplied to retrieve the next page of IDs. That ID is currently hardcoded to Long given NOMIS will typically use a Long when paging via an ID is used.


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/personalrelationships/ContactPersonReconciliationService.kt
```diff
@@ -52,128 +41,20 @@ class ContactPersonReconciliationService(
     const val TELEMETRY_PERSON_PREFIX = "contact-person-reconciliation"
   }
 
-  suspend fun generatePrisonerContactReconciliationReport(): PrisonerContactsReconciliationResult {
```
andymarke, line 55: all coroutine and channel stuff extracted to a new file

lbennett-moj (approved):

PR #1220, opened 2025-05-30
SDIT-2454: ✨ Listen to organisation events

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/organisations/OrganisationsService.kt
```diff
@@ -62,29 +66,77 @@ class OrganisationsService(
...
-        } ?: also {
-          telemetryClient.trackEvent("organisation-deleted-skipped", telemetryMap)
+      synchronise {
+        name = entityName
+        telemetryClient = this@OrganisationsService.telemetryClient
+        retryQueueService = organisationsRetryQueueService
+        eventTelemetry = telemetryMap
+
+        checkMappingDoesNotExist {
+          mappingApiService.getByDpsOrganisationIdOrNull(dpsOrganisationId)
```
petergphillips, line 84: technically I don't need to go to the mapping service for organisations, but seemed easier to do so to ensure that we don't call nomis prisoner api to do a create when it already exists etc.

PR #1232, opened 2025-06-03
SDIT-2744 Incidents whatChanged field is nullable

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incidents/IncidentsService.kt
```diff
@@ -46,7 +46,7 @@ data class IncidentAdditionalInformation(
   val reportReference: Long,
   val source: CreatingSystem,
   // TODO - check - we may not need this
-  val whatChanged: String,
+  val whatChanged: String?,
```
karenmillermoj, line 49: Will add test if using this field (otherwise will pull out at a later date)

PR #1233, opened 2025-06-03
SDIT-2744: ✨ Add in incidents mapping

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incidents/IncidentsMappingApiService.kt
```diff
@@ -0,0 +1,45 @@
...
+    .retrieve()
+    .awaitBody()
+
+  suspend fun createIncidentMapping(mappings: IncidentMappingDto) = webClient.post()
+    .uri("/mapping/incidents")
+    .bodyValue(mappings)
+    .retrieve()
+    .awaitBodilessEntityOrThrowOnConflict()
+
+  suspend fun deleteByNomisIncidentId(nomisIncidentId: Long) {
```
karenmillermoj, line 36: There is already a deleteByDpsIncidentId in the mapping service - the data from Dps will already have the dps id (can use either tbh)

PR #1237, opened 2025-06-04
SDIT-2484 reconciliation - adding sentences, just getting something w…

lbennett-moj:
…orking at this stage - more thought into what is essential and tests to come


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingReconciliationService.kt
```diff
@@ -127,6 +186,20 @@ class CourtSentencingReconciliationService(
         }
 
         differences.addAll(compareLists(sortedDpsAppearances, sortedNomisAppearances, "$parentProperty.appearances"))
+
```
andymarke, line 189: I'm not sure how that compareList works; I would have thought you would have two lists and you would check each item in one list matches an item in the other and visa versa. Therefore the order is irrelevant .

I'm just trying to understand how it copes with say 3 sentences all start on the same day with same offences code etc but different lengths

PR #1270, opened 2025-06-11
SDIT-2817 Set and Get breach appearance court Id for recalls

andymarke:
A new recall will create a breach court hearing; these will be stored in mapping and used for later deletes and updates

TODO - clean up recall mappings after a recall delete


petergphillips (approved):
LGTM

PR #1277, opened 2025-06-12
mh SDIT 2793 update and remove client tracking

mikehalmamoj:
- **SDIT-2793 Upgrade dependencies and use client tracking from library**
- **SDIT-2793 Upgrade gradle**
- **SDIT-2793 Use client tracking from library**


petergphillips (approved):
LGTM

PR #1280, opened 2025-06-12
SDIT-2817 Call revert recall when deleting and old recall

andymarke:
NOMIS needs to do something slightly different when reverting back to an old recall


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtCasesToNomisIntTest.kt
```diff
@@ -3176,15 +3176,15 @@ class CourtCasesToNomisIntTest : SqsIntegrationTestBase() {
           )
 
           courtSentencingMappingApi.stubGetAppearanceRecallMappings(
-            "dc71f3c5-70d4-4faf-a4a5-ff9662d5f714",
```
andymarke, line 3179: The major change in the NOMIS api, it needs to current court events to delete them

lbennett-moj (approved):

PR #1314, opened 2025-06-20
SDIT-2804 Protect endpoints called from cronjonbs

mikehalmamoj:


helm_deploy/hmpps-prisoner-to-nomis-update/templates/activities-tidy-mappings-cronjob.yaml
```diff
@@ -20,10 +20,37 @@ spec:
...
+              volumeMounts:
+                - name: hmpps-auth-token-requests
+                  mountPath: /bin/hmpps-auth-token-requests.sh
+                  readOnly: true
+                  subPath: hmpps-auth-token-requests.sh
               args:
                 - /bin/sh
                 - -c
-                - curl --retry 2 -XDELETE http://hmpps-prisoner-to-nomis-update/activities/mappings/unknown-mappings
+                - |
```
petergphillips, line 31: looks quite a long script here - would be better to maybe specify the URL and verb into a function as that looks like the only bits that are different between each snippet

petergphillips (dismissed):
LGTM

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #1334, opened 2025-06-27
SDIT-2821 Incidents reconciliation

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incidents/IncidentsReconciliationService.kt
```diff
@@ -0,0 +1,259 @@
+package uk.gov.justice.digital.hmpps.prisonertonomisupdate.incidents
+
+import com.microsoft.applicationinsights.TelemetryClient
+import kotlinx.coroutines.Dispatchers
```
karenmillermoj, line 4: There is still work to do in the incidents reconciliation service, but this is the main comparison - copied and changed slightly from the from-nomis service. Improvements to follow.

petergphillips (approved):
LGTM

PR #1381, opened 2025-07-09
SDIT-2759 Log missing location mapping and revert to deprecated value if does not exist

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -405,6 +418,14 @@ class AdjudicationsService(
     }
   }
 
+  // TODO future change Remove logging and use value only from mapping service once satisfied location picked up correctly
```
karenmillermoj, line 421: This is just a temporary change to log the data if there are any failures - once we are confident all is ok, or looking ok...I will remove references to the deprecated field

mikehalmamoj (approved):

PR #1384, opened 2025-07-10
SDIT-2517 map from dps location

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/activities/ActivitiesService.kt
```diff
@@ -152,16 +152,16 @@ class ActivitiesService(
     excludeBankHolidays = !runsOnBankHoliday,
     outsideWork = outsideWork,
     endDate = endDate,
-    internalLocationId = internalLocation?.id?.toLong(),
+    internalLocationId = internalLocation?.let { mappingService.getLocationMappingGivenDpsId(internalLocation.dpsLocationId!!).nomisLocationId },
```
mikehalmamoj, line 155: The main change is to take the DPS location ID from activities and map it to the NOMIS location ID via the mapping service.

The `dpsLocationId!!` is currently nullable to avoid a breaking change. Soon it will be non-nullable and `id` will be removed entirely.

PR #1470, opened 2025-08-05
SDIT-2759 Remove deprecated locationId by calling location mapping service instead

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsService.kt
```diff
@@ -418,14 +411,6 @@ class AdjudicationsService(
     }
   }
 
-  // TODO future change Remove logging and use value only from mapping service once satisfied location picked up correctly
```
karenmillermoj, line 421: Monitoring of this over the last month has shown only connectivity issues (i.e. no missing location ids) so can now be removed

PR #1490, opened 2025-08-13
SDIT-2982 Create court appearance now uses batch mappings endpoint

andymarke:
For the next phase the createAppearance NOMIS API can return the cloned court cases when appearance created on an old booking, these will be sent in the same payload for the mappings service. So this change should end up with the same behaviour except using a different mapping endpoint.

Also speeded up tests


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -1259,7 +1271,6 @@ class CourtSentencingService(
   }
 
   private inline fun <reified T> String.fromJson(): T = objectMapper.readValue(this)
-  private fun Any.toJson(): String = objectMapper.writeValueAsString(this)
```
andymarke, line 1262: unused

steverendell (approved):
lgtm

PR #1503, opened 2025-08-15
SDIT-2982 Send clone SQS notification after mappings created.

andymarke:
Cloning results in old cases and the new cases.
The mappings that pointed at the old cases are switched to the new cases on the latest booking.

The old cases are then sync'd back to DPS and new mappings are created. This must happen after the mappings are updated else we get a race condition.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
@@ -56,6 +56,7 @@ import uk.gov.justice.digital.hmpps.prisonertonomisupdate.services.synchronise
 import java.time.LocalDateTime
 import java.time.LocalTime
 import java.util.UUID
+import kotlin.collections.plus
 
 typealias MappingSentenceId = uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomismappings.model.SentenceId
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -196,24 +197,10 @@ class CourtSentencingService(
                 .also { telemetryMap["courtEventCharges"] = it.toString() },
             ),
           ).also { response ->
-            response.clonedCourtCases?.also { response ->
```
andymarke, line 199: moved to createAppearanceMappingsAndNotifyClonedCases()

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -225,27 +212,32 @@ class CourtSentencingService(
...
-      mappingType = CourtCaseBatchMappingDto.MappingType.DPS_CREATED,
-    ),
-    mappingsToUpdate = CourtCaseBatchUpdateMappingDto(
-      courtCases = toCourtCases(),
-      courtAppearances = toCourtAppearances(),
-      courtCharges = toCourtCharges(),
-      sentences = toSentences(),
-      sentenceTerms = toSentenceTerms(),
     ),
+    offenderNo = offenderNo,
```
andymarke, line 239: added this data just for the SQS message send

mikehalmamoj (approved):

PR #1558, opened 2025-08-29
SDIT-3009 Prep work for recalls on previous bookings

andymarke:
TODO - adding the booking clone response - for now assume it is always null

No real change other than we will send separate adjustment sync messages for each individual adjustment per sentence. This is because for clone cases as well as updating adjustments, we will be creating them which aren't idempotent so we need a message each. This has no functional effect at all (and typically a sentence won't have more than one adjustment anyway).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
@@ -710,12 +710,88 @@ class CourtSentencingService(
     }
   }
 
-  suspend fun createRecallAppearanceRetry(message: CreateMappingRetryMessage<CourtAppearanceRecallMappingsDto>) = courtCaseMappingService.createAppearanceRecallMappings(message.mapping).also {
-    telemetryClient.trackEvent(
-      "court-appearance-recall-create-mapping-retry-success",
-      message.telemetryAttributes,
-      null,
-    )
+  suspend fun tryCreateAppearanceMappingsAndNotifySentenceAdjustmentsAndClonedCases(mappingsWrapper: RecallAppearanceAndCreateMappingsWrapper, offenderNo: String, telemetry: Map<String, String>) {
+    try {
+      createAppearanceMappingsAndNotifySentenceAdjustmentsAndClonedCases(
+        mappingsWrapper = mappingsWrapper,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -1067,35 +1143,21 @@ class CourtSentencingService(
...
+              dpsRecallId = recallId,
+            ),
             offenderNo = offenderNo,
-            sentences = response.sentenceAdjustmentsActivated.map {
-              SentenceIdAndAdjustmentIds(
-                sentenceId = it.sentenceId,
-                adjustmentIds = it.adjustmentIds,
-              )
-            },
+            // TODO
```
andymarke, line 1154: once we set this we need a whole set of tests showing what happens when the case is cloned due to it being on a old booking

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtCasesToNomisIntTest.kt
```diff
@@ -3218,227 +3218,259 @@ class CourtCasesToNomisIntTest : SqsIntegrationTestBase() {
...
-              assertThat(it["dpsRecallId"]).isEqualTo("dc71f3c5-70d4-4faf-a4a5-ff9662d5f714")
-              assertThat(it["offenderNo"]).isEqualTo(OFFENDER_NO)
-              assertThat(it["dpsSentenceIds"]).isEqualTo("9ee21616-bbe4-4adc-b05e-c6e2a6a67cfc, 7ed5c261-9644-4516-9ab5-1b2cd48e6ca1")
-              assertThat(it["nomisSentenceSeq"]).isEqualTo("1, 2")
-              assertThat(it["nomisBookingId"]).isEqualTo("$BOOKING_ID")
-            },
-            isNull(),
-          )
+          @Test
+          fun `will send message to nomis migration for each sync sentence adjustments that have been activated`() {
```
andymarke, line 3355: only functional change is shown here;

previously we had two adjustments on same sentence generate two sync messages with both in the body of the message
now we send two messages with each adjustment in each message. The end result will be the same.

petergphillips (approved):
LGTM

PR #1560, opened 2025-08-29
SDIT-3009 Handle case cloning for recalls on previous bookings

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingService.kt
```diff
@@ -229,43 +231,36 @@ class CourtSentencingService(
         mappingType = CourtCaseBatchMappingDto.MappingType.DPS_CREATED,
       ),
       mappingsToUpdate = CourtCaseBatchUpdateMappingDto(
-        courtCases = toCourtCases(),
```
andymarke, line 232: refactor to allow reuse

PR #1609, opened 2025-09-18
SDIT-3051: ✨ Use generated api to call visit balance

petergphillips:


openapi-generator-ignore-visit-balance
```diff
@@ -0,0 +1,19 @@
...
+**/gradle/
+**/gradlew.bat
+**/README.md
+**/settings.gradle
+**/.openapi-generator-ignore
+**/client/infrastructure/
+**/src/test/
+**/api/*
+**/model/*
+!**/BalanceControllerApi.kt
```
petergphillips, line 13: we only generate the two endpoints and five model classes that we need, excluding all others

PR #1612, opened 2025-09-18
SDIT-2804 Try some reconciliation batch jobs in preprod

mikehalmamoj:
Not touched the existing cronjobs yet - the plan is to convert some to the new style of cronjob and run them in preprod only as a test (just after the existing cronjob finishes in preprod). If results are the same then we can get rid of the old cronjobs.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsReconciliationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsReconciliationService.kt
@@ -34,6 +34,25 @@ class AdjudicationsReconciliationService(
     val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
+  suspend fun generateAdjudicationsReconciliationReport() {
+    val activePrisonersCount = nomisApiService.getActivePrisoners(0, 1).totalElements
+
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/adjudications/AdjudicationsReconciliationService.kt
```diff
@@ -34,6 +34,25 @@ class AdjudicationsReconciliationService(
...
+        log.info("Adjudication reconciliation report completed with ${it.size} mismatches")
+        telemetryClient.trackEvent("adjudication-reports-reconciliation-report", mapOf("mismatch-count" to it.size.toString(), "success" to "true") + it.asMap())
+      }
+      .onFailure {
+        telemetryClient.trackEvent("adjudication-reports-reconciliation-report", mapOf("success" to "false"))
+        log.error("Adjudication reconciliation report failed", it)
+      }
+  }
+
+  private fun List<MismatchAdjudicationAdaPunishments>.asMap(): Map<String, String> = this.associate { it.prisonerId.offenderNo to ("${it.nomisAda.count}:${it.dpsAdas.count},${it.nomisAda.days}:${it.dpsAdas.days}") }
```
mikehalmamoj, line 54: The pattern seems to be copying the entire method from the existing Resource, but removing the special CoroutineScope `reportScope` (which was used to "fire and forget" but we don't want to forget about them any more).

PR #1615, opened 2025-09-19
SDIT-2804 Switch adjudications, alerts and appointments reconciliations to new cronjob template

mikehalmamoj:


steverendell (approved):
lgtm

PR #1654, opened 2025-10-01
SDIT-3037 Try exponential backoff retries

mikehalmamoj:


src/test/resources/application-test.yml
```diff
@@ -28,6 +28,7 @@ api.base.url:
   visit.balance: http://localhost:8101
 
 hmpps.sqs:
+  defaultErrorVisibilityTimeout: 0
```
mikehalmamoj, line 31: There were some projects relying on the default immediate retries so switching to the exponential backoff broke them. It seems sensible to make immediate retries the default for all tests, nobody seems to expect any different.

PR #2547, opened 2026-04-13
SDIT-3671 Prevent sync back to NOMIS of visit deletes

andymarke:
Normally sync'ng back to NOMIS a delete when a mapping still exists is fine. But for visits it isn't since the delete from DPS might be a NOMIS social visit that was converted from an Official Visit


steverendell (approved):
lgtm

PR #2584, opened 2026-04-20
SDIT-3660 Removed 409 create alert handling

andymarke:
We tried to create an alert mapping when a duplicate alert in NOMIS was detected. This would mean a double POST to NOMS API would be handled automatcially.

This caused more issues then it solved; since you can have the same alert code  in NOMIS multiple times but only one can be active.

This caused timing issues where an alert can be made inactive and the same alert code created around the same time. If the events are processed out of order the mappings ends up wrong.

Also refactored the webClient to use generated API.

Also refactored test to handle DLQ scenarios better

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsNomisApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsNomisApiService.kt
@@ -1,12 +1,10 @@
 package uk.gov.justice.digital.hmpps.prisonertonomisupdate.alerts
 
+import kotlinx.coroutines.reactor.awaitSingle
 import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClient
-import org.springframework.web.reactive.function.client.awaitBodilessEntity
-import org.springframework.web.reactive.function.client.awaitBody
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomisprisoner.api.AlertsResourceApi
-import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomisprisoner.model.AlertId
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomisprisoner.model.AlertResponse
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomisprisoner.model.CreateAlertCode
 import uk.gov.justice.digital.hmpps.prisonertonomisupdate.nomisprisoner.model.CreateAlertRequest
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsNomisApiService.kt
```diff
@@ -26,107 +22,57 @@ class AlertsNomisApiService(@Qualifier("nomisApiWebClient") private val webClien
...
     nomisAlert: CreateAlertRequest,
-  ): SuccessOrDuplicate<CreateAlertResponse, AlertId> = with(api) {
-    awaitSuccessOrDuplicate(
-      createAlertRequestConfig(
-        offenderNo = offenderNo,
-        createAlertRequest = nomisAlert,
-      ),
-    )
-  }
+  ): CreateAlertResponse = api.createAlert(
```
andymarke, line 25: No special 409 handling - just treat the same as every other error

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsNomisApiService.kt
```diff
@@ -26,107 +22,57 @@ class AlertsNomisApiService(@Qualifier("nomisApiWebClient") private val webClien
...
 
-  suspend fun updateAlert(bookingId: Long, alertSequence: Long, nomisAlert: UpdateAlertRequest): AlertResponse = webClient.put().uri(
-    "/prisoners/booking-id/{bookingId}/alerts/{alertSequence}",
-    bookingId,
-    alertSequence,
-  )
-    .bodyValue(nomisAlert)
-    .retrieve()
-    .awaitBody()
+  suspend fun updateAlert(bookingId: Long, alertSequence: Long, nomisAlert: UpdateAlertRequest): AlertResponse = api.updateAlert(
```
andymarke, line 30: Unrelated- use generated API methods

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/alerts/AlertsToNomisIntTest.kt
```diff
@@ -197,47 +192,36 @@ class AlertsToNomisIntTest : SqsIntegrationTestBase() {
           fun setUp() {
             alertsMappingApi.stubPostMappingFollowedBySuccess(HttpStatus.INTERNAL_SERVER_ERROR)
             publishCreateAlertDomainEvent(offenderNo = offenderNo, alertUuid = dpsAlertId)
+            waitForAnyProcessingToComplete("alert-create-success")
           }
```
andymarke, line 196: added waits in setup based on custom event rather than  `await untilAsserted` previously used

steverendell (approved):
lgtm

PR #2610, opened 2026-04-24
SDIT-3704 Allow null start dates for key dates

andymarke:
Some key date adjustments - i.e. SREMs have no start date - so just default to adjustment date


petergphillips (approved):
LGTM

PR #2655, opened 2026-05-05
SDIT-3775 Active visits was missing friday

andymarke:
Eventually this will run daily and then the full reconciliation will be weekly on Saturday


steverendell (approved):
lgtm

PR #2679, opened 2026-05-12
SDIT-3702: ✨ Add in cron prefix override

petergphillips:


helm_deploy/hmpps-prisoner-to-nomis-update/values.yaml
```diff
@@ -1,6 +1,8 @@
 generic-service:
   nameOverride: hmpps-prisoner-to-nomis-update
   productId: DPS060
+  cronPrefixName: to-nomis
+  cronPrefixName: hptonu
```
steverendell: This doesnt look right

steverendell (approved):

PR #2711, opened 2026-05-19
sdit_3818 adding mapping data to single offender recon

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/courtsentencing/CourtSentencingReconciliationService.kt
```diff
@@ -135,12 +136,14 @@ class CourtSentencingReconciliationService(
     offenderNo: String,
     nomisCaseIds: List<Long>,
     telemetryPrefix: String? = TELEMETRY_COURT_CASE_PRISONER_PREFIX,
+    extraInfo: Boolean? = false,
```
lbennett-moj, line 139: not the best name - will see how this evolves and maybe rename

andymarke (approved):

PR #2740, opened 2026-05-26
SDIT-3826 Handle breach appearance created by recall update

andymarke:
When a DPS recall is updated it may contain additional court cases. These will each be assigned a breach court appearance, so this changes ensures the mapping between the recall id and the hearings is correctly updated.

(always sends message back to migration service for the new court appearances for the sync back to DPS)


petergphillips (approved):
LGTM

PR #2797, opened 2026-06-03
SDIT-3855: 👽️ Switch to new incentives prisoner endpoint

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesDataRepairResource.kt
petergphillips, line 28: this is a bit naff, it is relying on the `bookingId` as returned from the incentive service

karenmillermoj (approved):
lgtm

PR #2853, opened 2026-06-17
SDIT-3773 Add staff reconciliation batch and endpoint

karenmillermoj:


helm_deploy/values-dev.yaml
```diff
@@ -95,6 +81,24 @@ generic-service:
       type: COURT_SCHEDULER_ALL_RECON
       schedule: "00 4 * * *"
       suspend: true
+    - name: prison-bal-recon
```
petergphillips, line 84: can be `prison-balance-recon` these days as the name limit is a lot longer

petergphillips (approved):
LGTM

PR #2892, opened 2026-06-25
SDIT-3714 Fix bug reporting IDs without reconciliation mismatches

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/movements/ExternalMovementsHelper.kt
```diff
@@ -20,6 +20,6 @@ internal fun <SOURCE, TARGET> findMatches(
   .filter { (_, trg) -> trg != null }
   .map { (src, trg) -> src to trg!! }
 
-class NomisMovementId(val bookingId: Long, val sequence: Int) {
+data class NomisMovementId(val bookingId: Long, val sequence: Int) {
```
mikehalmamoj, line 23: When calculating which IDs were missing, lookups in lists of this type were failing because equals wasn't implemented correctly.

andymarke (approved):

PR #2952, opened 2026-07-13
SDIT-3993: ✨ Add in list of bookings to ignore for incentive reconciliation

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesReconciliationService.kt
```diff
@@ -74,20 +82,48 @@ class IncentivesReconciliationService(
...
+              "bookingId" to mismatch.prisonerId.bookingId.toString(),
+              "nomisIncentiveLevel" to (mismatch.nomisIncentiveLevel ?: "null"),
+              "dpsIncentiveLevel" to (mismatch.dpsIncentiveLevel ?: "null"),
+            ),
+          )
+        }
+      } else {
         telemetryClient.trackEvent(
-          "incentives-reports-reconciliation-mismatch",
+          "incentives-reports-reconciliation-excluded-booking",
```
petergphillips, line 106: we alert on all events starting with `incentives-reports-reconciliation-mismatch` so this won't be alerted on

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/incentives/IncentivesReconciliationService.kt
```diff
@@ -74,20 +82,48 @@ class IncentivesReconciliationService(
...
+            "bookingId" to prisonerId.bookingId.toString(),
+            "reason" to "Excluding reconciliation mismatches for booking id ${prisonerId.bookingId}",
           ),
         )
+        null
       }
     } else {
+      if (excludedBooking) {
+        telemetryClient.trackEvent(
+          "incentives-reports-reconciliation-mismatch-excluded-booking-resolved",
```
petergphillips, line 118: but the resolved event will be (as starts with `incentives-reports-reconciliation-mismatch`)

mikehalmamoj (approved):

